PR #170, opened 2018-07-12
WIP: Attempt to fix Issue #109. Tests have not been modified. Extends polkadot-js/apps PR #137

ltfschoen:
Request review/feedback

### DIFF
--- a/packages/type-params/src/decode/index.ts
+++ b/packages/type-params/src/decode/index.ts
@@ -4,6 +4,7 @@
 
 import { EncodingVersions, Param$Decoded, Param$Type$Array, Param$Types, Param$Value, Param$Value$Array } from '../types';
 
+import encodeAddress from '@polkadot/util-keyring/address/encode';
 import u8aToBn from '@polkadot/util/u8a/toBn';
 
 import decodeValue from './value';
@@ -33,6 +34,10 @@ function decodeArray ([ type ]: Param$Type$Array, input: Uint8Array, version: En
 
   for (let index = 0; index < arrayLength; index++) {
     const decoded = decode(type, input.subarray(length), version);
+
+    if (type === 'AccountId') {
+      decoded.value = encodeAddress((decoded.value as Uint8Array))
+    }
 
     // NOTE small TS hack, since we don't have recursive arrays, assume primitives to push
     value.push(decoded.value as Param$Value);
--- a/packages/type-params/src/decode/value/account.ts
+++ b/packages/type-params/src/decode/value/account.ts
@@ -1,0 +1,15 @@
+// Copyright 2017-2018 @polkadot/params authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Param$Decoded } from '../../types';
+
+import encodeAddress from '@polkadot/util-keyring/address/encode';
+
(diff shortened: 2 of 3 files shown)


packages/type-params/src/decode/index.ts
```diff
@@ -34,6 +35,10 @@ function decodeArray ([ type ]: Param$Type$Array, input: Uint8Array, version: En
   for (let index = 0; index < arrayLength; index++) {
     const decoded = decode(type, input.subarray(length), version);
 
+    if (type === 'AccountId') {
```
jacogr: I cannot be here. All type decoding is done is `decodeValue`. As it stands we now have a situation where exactly the same logic exists in 2 places with no benefit. 

Array will always call through to the base and the base will always end up in decodeValue.

packages/type-params/src/decode/value/account.ts
```diff
@@ -0,0 +1,15 @@
...
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Param$Decoded } from '../../types';
+
+import encodeAddress from '@polkadot/util-keyring/address/encode';
+
+export default function account (input: Uint8Array, bitLength: 256 | 512): Param$Decoded {
+  const length = (bitLength / 8);
+  return {
```
jacogr: Spacing before.

packages/type-params/src/decode/value/account.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2017-2018 @polkadot/params authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { Param$Decoded } from '../../types';
+
+import encodeAddress from '@polkadot/util-keyring/address/encode';
+
+export default function account (input: Uint8Array, bitLength: 256 | 512): Param$Decoded {
```
jacogr: This will work for poc-1, but not for master - in master values that come via anything that is not storage has a prefix. (Hence if you look at the `decodeValue` implementation as it currently stands, we actually pass an offset through when we convert to u8a)

So basically -

1. additional parameter to be passed that identifies the type (`version: EncodingVersions`)
2. Here we do the have `const offset = version === 'poc-1' ? 0 : 1`
3. Assume, for now (will have to be fixed once we get enumerated accounts flowing), that we only have publicKey, not 8 or 32-bit enumerated accounts
4. `const publicKey = input.subarray(offset)`
5. `return { length: publicKey + offset,  value: encodeAddress(publicKey) }`

jacogr: Why are we passing in `bitLength`?

packages/type-params/src/decode/value/index.ts
```diff
@@ -27,7 +29,7 @@ export default function decodeValue (decode: Decoder, type: Param$Type, input: U
     switch (type) {
       // TODO Pass back the actual address, not publicKey?
       case 'AccountId':
-        return u8a(input, 256, version === 'poc-1' ? 0 : 1);
+        return account(input.slice(0, 32), 256);
```
jacogr: Changes as described above

1. pass in version
2. no need to pass in bitLength

jacogr:
Re-open if still applicable in /api, types moved in https://github.com/polkadot-js/api/pull/125

Issue #158, opened 2018-06-30
Convert primitives to object-with-baseclass


jacogr:
This one is a bit involved, so it probably makes sense starting with 1, seeing how it goes (implementing across the tree, i.e. here, api & apps) and then taking it further.

Basically, atm any value actually refers to the base type. For this example, let's take `BlockNumber`. In effect, it is `type BlockNumber = BN`, the idea is -

```
// probably may not be needed to pass generic, probably just Uint8Array (sliceables)
interface ParamEncoder <S> {
  encode (): S;
  toJSON (): any;
}

interface ParamDecoder <C, O> {
  decode (input: S): C;
  fromJSON (input: any): C;
}

interface BaseParam <T> { // or BasePrimitive, or BaseType, or...
  value: T;
}

class BlockNumber implements ParamEncoder<Uint8Array>, BaseParam<BN> {
  value: BN;

  encode (): Uint8Array {
    // ... convert to sliceable
  }

  toJSON (): any {
    // yes, we have utils for this, just a sample
    return '0x' + this.value.toString(16);
  }

  static decode (input: Uint8Array): BlockNumber {
    return new BlockNumber(...);
  } 

  static fromJSON (input: any): BlockNumber {
    return new BlockNumber(...);
  }
}  
```

The benefits we get -

1. consistent interface for all values
2. encoding/decoding is not done in different places across the codebase
3. instead of `import encoder from '@polkadot/primitives/bn/encoder'; enoder(bn)` we just do `bn.encode()`

BlockNumber here is actually an interesting one -

1. It may encode differently depending on chain (sizes, Nonce & Balance are currently good examples of this)
2. It shares the same type with others, only differring in length, i.e. Balance & Nonce are also with the same logic.
3. With the above in mind, it probably makes a bit more sense doing Nonce or Balance first (the latter has further-reaching UI impacts, so Nonce conevrs all the bases, but has a lesser footprint)

jacogr:
This is a bit of a monster, even just to get started. So PR #172 started, need to decide -

1. How to pass through the encoding size (could be 32 or 64 here, depending on chain/impl)
2. How to handle storage with that (not an issue here, but will be AccountId where the accounts are enumerated)

So... structure is there about more-or-less at this point what is expected, but need to make sure we don't paint ourselves into a corner.

jacogr:
Closing done in API https://github.com/polkadot-js/api/issues/161

PR #172, opened 2018-07-13
Primitives as classes (AccountIndex)

jacogr:
Start of https://github.com/polkadot-js/common/issues/158

### DIFF
--- a/packages/type-primitives/src/AccountIndex.ts
+++ b/packages/type-primitives/src/AccountIndex.ts
@@ -1,0 +1,34 @@
+// Copyright 2017-2018 @polkadot/primitives authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { AccountIndex, AccountIndexDecoder } from './types';
+
+import BN from 'bn.js';
+import bnToBn from '@polkadot/util/bn/toBn';
+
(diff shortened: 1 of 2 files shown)


jacogr:
Types moved in https://github.com/polkadot-js/common/pull/191

packages/type-primitives/src/AccountIndex.ts
```diff
@@ -0,0 +1,34 @@
+// Copyright 2017-2018 @polkadot/primitives authors & contributors
+// This software may be modified and distributed under the terms
+// of the ISC license. See the LICENSE file for details.
+
+import { AccountIndex, AccountIndexDecoder } from './types';
+
+import BN from 'bn.js';
+import bnToBn from '@polkadot/util/bn/toBn';
+
+class AccountIndexImpl implements AccountIndex {
```
jacogr: This needs to probably extends a "Number" class that has all the basics - only things different between AccountIndex, BlockNumber, Balance, u32, u64, etc. is the actual sizes.

so ...

```
class Number {
  constructor (bitLength: number = 32 | 64 | 128 | 256) {
    this._bitLength = bitLength;
  }
...
}

class AccountIndex extends Number {
  constructor () {
    super(64);
  }
...
}
```

PR #208, opened 2018-09-02
Add RocksDb interface

jacogr:


jacogr:
Let the chips fall where they may, let's see if LevelDb is fixed and working and test the RocksDb as part of an actual implementation... tests here only tell half the story it would seem.

PR #215, opened 2018-09-11
Classify Keyring

jacogr:
Aligns and simplifies use of Keyring with the rest of the interfaces where closures have been converted to ES6 objects. (e.g. large parts of the API)

Going from 

```
import createKeyring from '@polkadot/util-keyring';

const keyring = createKeyring();
const pair = keyring.addFromSeed(<some_seed>);
const signature = pair.sign(<some_message>);
```

to new version -

```
import Keyring from '@polkadot/util-keyring';

const keyring = new Keyring();
const pair = keyring.addFromSeed(<some_seed>);
const signature = pair.sign(<some_message>);
```

There will be work on the UI side now as well to implement this is ui-keyring. (Probably inherit from object and add the new methods?)

cc @chevdor 
cc @ltfschoen

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -9,5 +9,5 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.29.33"
+  "version": "0.30.0"
 }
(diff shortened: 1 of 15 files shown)


jacogr:
Does not actually seem to be impacts on the UI side of thing since it pulls in testKeyring (which had no interface change). So no impact in creation.

PR #246, opened 2018-11-12
Explicit Array<any>

jacogr:
Reported by @chevdor on using the libraries via npm (with a different TS setup)

In short, where we only do `const something = []` TS _sometimes_ (depending on config) would not see those as `any[]` but rather `never[]`. So be explicit where we define arrays.

### DIFF
--- a/packages/util-rlp/src/decoder/listShort.ts
+++ b/packages/util-rlp/src/decoder/listShort.ts
@@ -5,9 +5,9 @@
 import { DecodeFunc, DecodeOutput } from './types';
 
 export default function decodeListShort (decode: DecodeFunc, input: Uint8Array): DecodeOutput {
+  const decoded: Array<any | Uint8Array> = [];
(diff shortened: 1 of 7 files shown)


packages/util-rlp/src/decoder/listShort.ts
```diff
@@ -5,9 +5,9 @@
 import { DecodeFunc, DecodeOutput } from './types';
 
 export default function decodeListShort (decode: DecodeFunc, input: Uint8Array): DecodeOutput {
+  const decoded: Array<any | Uint8Array> = [];
```
jacogr: This is not "quite" correct - any includes Uint8Array anyway. But it gives a tiny bit of intent. (RLP will be removed shortly as it stands, PoC-3 have gotten rid of it completely).

(Funnily this was not one of the issues raised, they are all in trie-hash, however the same issue would manifest here)

PR #251, opened 2018-11-14
Allow blake2 & parity-codec in trie

jacogr:
Requires/contains https://github.com/polkadot-js/common/pull/256


packages/trie-codec/src/NodeHeader.ts
```diff
@@ -0,0 +1,173 @@
...
+    throw new Error('Unreachable');
+  }
+
+  get encodedLength (): number {
+    const nodeType = this.nodeType;
+
+    if (nodeType === NODE_TYPE_NULL || nodeType === NODE_TYPE_BRANCH) {
+      return 1;
+    } else if (nodeType === NODE_TYPE_EXT) {
+      const nibbleCount = (this.value.raw as ExtensionHeader).toNumber();
```
amaurymartiny: `this.value` is an ExtensionHeader already, I think you don't need raw here

packages/chainspec/LICENSE
```diff
@@ -0,0 +1,201 @@
+                              Apache License
```
jacogr: Package created so there is a common place where the chainspecs can be held. Currently these are actually included in the client. Putting it here (and they don't update frequently) has the added benefit that we can actually use actual data in tests, for instance to calculate the trie-roots and then compare against known.

packages/trie-hash/test/data.ts
```diff
@@ -0,0 +1,263 @@
+// Copyright 2017-2018 @polkadot/trie-hash authors & contributors
```
jacogr: These are basically the Rust tests - as a packaged file so we can use it both in trie-hash and trie-db to check the results

packages/util/src/compact/addLength.spec.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2017-2018 @polkadot/util authors & contributors
```
jacogr: types/codec/Compact.statics follow - api/types to be updated to use these

packages/util/src/u8a/toU8a.spec.js
```diff
@@ -33,6 +33,14 @@ describe('u8aToU8a', () => {
     );
   });
 
+  it('creates from Array', () => {
```
jacogr: It was uncovered... weak moment for my OCD.

PR #280, opened 2018-12-20
Fix negative BNs

amaurymartiny:
- Fix some bugs
- Also adds encoding for BN to hex/u8a with 2's complement


packages/util-rlp/test/getTests.ts
```diff
@@ -14,7 +14,7 @@ export default function getTests (file: string) {
     name,
     input: tests[name].in[0] !== '#'
       ? tests[name].in
-      : bnToU8a(new BN(tests[name].in.slice(1)), -1, true),
+      : bnToU8a(new BN(tests[name].in.slice(1)), { isLe: false, isNegative: false }),
```
amaurymartiny: This one bugged me for 1h, I thought I f*cked up somewhere... until I check on the internet, Ethereum uses BE.

jacogr: 100%

packages/util/src/bn/toHex.ts
```diff
@@ -25,10 +31,20 @@ const ZERO_STR = '0x00';
...
   if (!value) {
     return ZERO_STR;
   }
 
-  return hexFixLength(bnToBn(value).toString(16), bitLength, true);
+  const _options = {
+    isLe: false,
+    isNegative: false,
+    // Backwards-compatibility
+    ...(isNumber(options) ? { bitLength: options } : options)
```
jacogr: This made me frown, however we have existing tests to cover it, so all ok.

packages/util/src/bn/toU8a.ts
```diff
@@ -23,25 +27,36 @@ import bnToHex from './toHex';
  * bnToU8a(new BN(0x1234)); // => [0x12, 0x34]
  * ```
  */
-export default function bnToU8a (value: BN | number, bitLength: number, isLe: boolean): Uint8Array {
-  const byteLength = Math.ceil(bitLength / 8);
+export default function bnToU8a (value: BN | number | null, options?: Options): Uint8Array;
```
jacogr: We will need to come back and only keep this - first I will go through the code and cleanup to new version, then we can drop. It just becomes messy. (I am 100% with you that iw was needed, we just need to cleanup ASAP - will log and start when bored)

PR #306, opened 2019-02-15
Cleanup circular deps

jacogr:
Fix for circular dep in https://github.com/paritytech/schnorrkel-js/issues/7 (no async fix yet, up next)


jacogr:
Split wasm part into `schnorrkel_js_wasm` which is basically 

```js
module.exports = Buffer.from('...', 'base64');
```

Simplify creation part with function that takes imports as an arg

```js
const bytes = require('./schnorrkel_js_wasm');

module.exports = function init (schnorrkel_js) {
  const imports = { './schnorrkel_js': schnorrkel_js };
  const wasmModule = new WebAssembly.Module(bytes);
  const wasmInstance = new WebAssembly.Instance(wasmModule, imports);

  return wasmInstance.exports;
};
```

PR #359, opened 2019-03-11
Parsing function for JSON objects

Stefie:
Checks to see if the input value is a valid JSON object.
 It returns false if the input is JSON parsable, but not an Javascript object.

Use-cases:
- To validate if uploaded files contain valid JSON objects
- To validate if manual code input in editors contains valid JSON objects

### DIFF
--- a/packages/util/src/is/jsonObject.spec.ts
+++ b/packages/util/src/is/jsonObject.spec.ts
@@ -1,0 +1,51 @@
+// Copyright 2017-2019 @polkadot/util authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { isJsonObject } from './index';
+
+const jsonObject = `{
+    "Test": "1234",
+    "NestedTest": {
+      "Test": "5678"
+    }
+  }`;
+
+describe('isJsonObject', () => {
+
+  it('returns true on empty objects', () => {
+    expect(
+      isJsonObject({})
+    ).toEqual(true);
+  });
+
+  it('returns true on JSON objects', () => {
+    expect(
+      isJsonObject(jsonObject)
+    ).toEqual(true);
+  });
+
+  it('returns false on valid JSON but typeof number', () => {
+    expect(
+      isJsonObject(1234)
+    ).toEqual(false);
+  });
+
+  it('returns false on valid JSON but null', () => {
+    expect(
+      isJsonObject(null)
+    ).toEqual(false);
+  });
+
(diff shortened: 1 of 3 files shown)


packages/util/src/is/jsonObject.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const jsonObject = `{
+    "Test": "1234",
+    "NestedTest": {
+      "Test": "5678"
+    }
+  }`;
+
+describe('isJsonObject', () => {
+
```
jacogr: (nitpick) remove extra blank line

packages/util/src/is/jsonObject.ts
```diff
@@ -0,0 +1,45 @@
...
+ *   "Test": "5678"
+ *  }
+ * }); // => true
+ * isJsonObject(1234); // JSON parsable, but not an object =>  false
+ * isJsonObject(null); // JSON parsable, but not an object => false
+ * isJsonObject('not an object'); // => false
+ * ```
+ */
+export default function isJsonObject (value: any): value is ObjectIndexed {
+
```
jacogr: (nitpick) extra blank line

packages/util/src/is/jsonObject.ts
```diff
@@ -0,0 +1,45 @@
...
+  value = typeof value !== 'string'
+  ? JSON.stringify(value)
```
jacogr: Weird indentation, prefer

```js
value = typeof value !== 'string'
  ? JSON.stringify(value)
  : value;
```

packages/util/src/is/jsonObject.ts
```diff
@@ -0,0 +1,45 @@
...
+  : value;
+
+  try {
+    value = JSON.parse(value);
+    return (typeof value === 'object' && value !== null);
```
jacogr: (nitpick) extra blank line to split blocks. (return and parsing)

packages/util/src/is/jsonObject.ts
Stefie: done.

packages/util/src/is/jsonObject.spec.ts
Stefie: done.

packages/util/src/is/jsonObject.ts
Stefie: true, I changed it.

packages/util/src/is/jsonObject.ts
Stefie: changed.

PR #384, opened 2019-03-26
Add singleton package assertion (& add to keyring)

jacogr:
For https://github.com/polkadot-js/api/issues/795


amaurymartiny:
Note: could add a quick test: calling this function twice should error

packages/util/src/assertSingletonPackage.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+describe('assertSingletonPackage', () => {
+  const NAME = 'assertSingletonPackage';
+
+  it('should not throw the first time', () => {
+    expect(
+      () => assertSingletonPackage(NAME)
+    ).not.toThrow();
+  });
+
+  it('should throw the second time', () => {
```
xlc: this could be a flaky test as jest may run tests in parallel or out of order

PR #401, opened 2019-05-07
Add BN min/max functions

kwingram25:


packages/util/src/bn/min.ts
```diff
@@ -0,0 +1,30 @@
...
+export default function min (
+  ...items: BN[]
+): BN | null {
+  if (!items || !items.length) {
+    return null;
+  }
+
+  return items.reduce((acc: BN, val: BN) => {
+    return (val.lt(acc) ? val : acc);
+  }, new BN('1e500'));
```
amaurymartiny: You could use `items[0]` here

packages/util/src/bn/min.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export default function min (
+  ...items: BN[]
+): BN {
+  assert(
+    items && items.length >= 2,
+    'Must provide two or more BN arguments'
+  );
+
+  return items.reduce((acc: BN, val: BN) => BN.min(acc, val), new BN('1e500'));
```
amaurymartiny: Can use item[0], so that you don't hardcode a max value

packages/util/src/bn/min.ts
```diff
@@ -0,0 +1,31 @@
...
+ * min([new BN(1), new BN(3), new BN(2)]).toString(); // => '1'
+ * ```
+ */
+export default function min (
+  ...items: BN[]
+): BN {
+  assert(
+    items && items.length >= 2,
+    'Must provide two or more BN arguments'
+  );
```
amaurymartiny: This assert is useful, but (totally optional), we could add something like https://stackoverflow.com/questions/49910889/typescript-array-with-minimum-length so that the compiler will also throw an error when `bnMin()` or `bnMin(a)`

packages/util/src/bn/max.ts
```diff
@@ -20,12 +20,7 @@ import assert from '../assert';
  * ```
  */
 export default function max (
-  ...items: BN[]
+  ...items: BnList
 ): BN {
-  assert(
```
jacogr: Don't drop the assert, use both :)

amaurymartiny: Yeah, my message was maybe not clear. Assert is useful, BNList type is optional

PR #416, opened 2019-06-04
Move address helpers to util-crypto

jacogr:
Move base implementation for decodeAddress, encodeAddress & setAddressPrefix to `@polkadot/util-crypto` (since it deals with hashing, it is relevant despite the name)

External interfaces are not affected, i.e. the exports on keyring stays as-is. What this does allow us to do however is to depend on `util-crypto` (already-existing dependency) in `@polkadot/api` only. This removes a lot of the complications from using the API and having duplicates appear.

Deps flow:

```
util -> util-crypto -> keyring
                    -> api
```

### DIFF
--- a/packages/util/src/address/sshash.ts
+++ b/packages/util/src/address/sshash.ts
@@ -1,9 +1,12 @@
-// Copyright 2017-2019 @polkadot/keyring authors & contributors
+// Copyright 2017-2019 @polkadot/util authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { stringToU8a, u8aConcat } from '@polkadot/util';
+// This is not great - here we now have a dependency from util -> crypto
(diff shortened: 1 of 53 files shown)


packages/util/src/address/sshash.ts
```diff
@@ -1,10 +1,13 @@
-// Copyright 2017-2019 @polkadot/keyring authors & contributors
+// Copyright 2017-2019 @polkadot/util authors & contributors
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { stringToU8a, u8aConcat } from '@polkadot/util';
+// This is not great - here we now have a dependency from util -> crypto
 import { blake2AsU8a } from '@polkadot/util-crypto';
```
jacogr: This is the part that causes all hell to break loose (Comment actually made before seeing the failures)

Only solution would be to move the address/* helpers to util-crypto

jacogr:
Follow-up with #420

PR #419, opened 2019-06-11
Simplify @polkadot/keyring KeyringPair

axelchalon:
Modify the KeyringPair interface to be consistent with [changes](https://github.com/polkadot-js/ui/pull/144) made to `@polkadot/ui-keyring`

### DIFF
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -89,7 +89,9 @@ export default function createPair (type: KeypairType, { publicKey, secretKey }:
     lock: (): void => {
       secretKey = new Uint8Array(0);
     },
-    publicKey,
+    get publicKey () {
+      return publicKey;
+    },
     setMeta: (additional: KeyringPair$Meta): void => {
       meta = { ...meta, ...additional };
     },
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -47,7 +47,7 @@ const verify = (type: KeypairType, message: Uint8Array, signature: Uint8Array, p
  * - `encodePkcs8` function when provided with the correct passphrase associated with the account pair
  * and when the secret key is in memory (when the account pair is not locked) it returns an encoded
  * public key of the account.
- * - `getMeta` returns the metadata that is stored in the state of the pair, either when it was originally
+ * - `meta` is the metadata that is stored in the state of the pair, either when it was originally
  * created or set via `setMeta`.
  * - `publicKey` returns the public key stored in memory for the pair.
  * - `sign` may be used to return a signature by signing a provided message with the secret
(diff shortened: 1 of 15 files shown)


packages/keyring/src/pair/index.ts
```diff
@@ -65,8 +65,7 @@ const verify = (type: KeypairType, message: Uint8Array, signature: Uint8Array, p
 export default function createPair (type: KeypairType, { publicKey, secretKey }: PairInfo, meta: KeyringPair$Meta = {}, encoded: Uint8Array | null = null): KeyringPair {
   return {
     type,
-    address: (): string =>
-      encodeAddress(publicKey),
+    address: encodeAddress(publicKey),
```
jacogr: There is an issue with this - if the prefix changes, this is not taken into account for the address. SWo basically, depending on where the set is called by the user, if may be incorrect.

packages/keyring/src/pair/index.ts
```diff
@@ -82,15 +81,14 @@ export default function createPair (type: KeypairType, { publicKey, secretKey }:
...
+    get meta () {
+      return meta;
+    },
+    isLocked: (!secretKey || secretKey.length === 0),
     lock: (): void => {
       secretKey = new Uint8Array(0);
     },
-    publicKey: (): Uint8Array =>
-      publicKey,
+    publicKey,
```
jacogr: The same as above applies to the publicKey - this is set as we decode (so you could instantiate with an empty value, and unlock/decode to extract).

packages/keyring/src/pair/index.ts
jacogr: Think I am actually wrong here - passed by ref, so changes should reflect. Would probably be good adding an explicit test around this then. 

1.  Get address, compare to known 42 prefix 
2. Set prefix to 68
3. Compare to known 68 prefix

packages/keyring/src/pair/index.ts
jacogr: Probably also incorrect...

packages/keyring/src/pair/index.ts
jacogr: So this is the test that (is not available, but would worin in master, but doesn't after this -

```js
// in pair/index.spec.ts

// once master is merged this import would be @polkadot/util-crypto
import { setAddressPrefix } from '../address';

describe('pair', () => {
  ...
  it('allows encoding of address with different prefixes', () => {
    expect(keyring.alice.address).toEqual(
      '5GoKvZWG5ZPYL1WUovuHW3zJBWBP5eT8CbqjdRY4Q6iMaQua'
    );

    setAddressPrefix(68);

    expect(keyring.alice.address).toEqual(
      '7sGUeMak588SPY2YMmmuKUuLz7u2WQpf74F9dCFtSLB2td9d'
    );
  });
```

packages/keyring/src/pair/index.ts
axelchalon: I think this is still an issue; verifying.

jacogr: Yes, also wanted to check that one - probably best just making it a getter anyway, that way we know. (As per my comments, when not completely awake, I do sometimes miss the plot as to when things are by value vs by ref, at least a getter won't be surprising as to intent)

PR #424, opened 2019-06-14
Add missing dependencies to util-crypto

Stefie:
Util-crypto currently breaks the JS API because `bs58` is being imported but missing in the dependencies and therefore not installed.


jacogr:
https://github.com/polkadot-js/common/blob/master/yarn.lock#L1870
https://github.com/polkadot-js/common/blob/master/yarn.lock#L3167

jacogr:
The correct fix is moving it from `@polkadot/keyring` to `@polkadot/util-crypto`

Stefie:
I didn't check the other packages for unused imports. I removed it from `keyring` now

Stefie:
GitHub notification test for @shawntabrizi

PR #435, opened 2019-07-10
Swap to using eslint

jacogr:


packages/keyring/src/index.spec.ts
```diff
@@ -137,7 +137,7 @@ describe('keypair', () => {
       );
     });
 
-    it('signs and verifies', () => {
+    it('signs and verifies', (): void => {
```
amaurymartiny: hmm i'm not sure i like this, it adds a lot of useless code (TS type inference works well)

jacogr: It is part of the default/recommended TS rules. So stuck with it.

jacogr: "Only" made 2 adjustments to the recommended -

- 2 spaces instead of the 4
- don't require "public" for constructors (I honestly didn't want to, but just cannot wrap my head around `public constructor` everywhere - happy to revert that though, I can adapt)

https://github.com/polkadot-js/dev/blob/master/packages/dev/config/eslint.js#L17

jacogr: Yes, I should revert the second. Just need to adapt, otherwise making up random rules.

amaurymartiny: oh well, let's adapt then

jacogr: Yeap, change the constructor thing as well, so now we have stuff like https://github.com/polkadot-js/common/pull/435/files#diff-7ead68f60a93c3658858fa52aad2827cR65

PR #440, opened 2019-07-13
added check if valid address to docs

JesseAbram:
Let me know if this should be in the utils folder


jacogr:
Thanks a million.

I think the location is perfect, keyring makes sense wince it is where people typically look to manage addresses.

PR #450, opened 2019-07-30
Handing high decimals scenes

WoeOm:
when a number decimals is too high, const mid will be a negative.

```
  it('formats 123,456,789,000 (decimals=36)', (): void => {
    expect(
      formatBalance(TESTVAL, true, 36)
    ).toEqual('0.123y Unit');
  });

```


jacogr:
Thank you.

PR #459, opened 2019-08-24
fix: 123m Unit => 123 mUnit

yjkimjunior:


jacogr:
Sadly the display this way around does not work in the polkadot-is apps UI at least. We had this right at the start, it ends up being a run of characters.  So not going to merge as a “fix”.

yjkimjunior:
:( why does it end up a run of characters? the `100m DOT` wouldn't just become `100 mDOT`?

jacogr:
Look at the example as provided above, ie the way the formatting actually started off as. 

Other options considered “1.234 M KSM” (lotsa space, most readable), “1.234MKSM” (worst of the lot)

PR #468, opened 2019-09-10
Rename keyring constructor parameter type -> defaultType

JoshOrndorff:
Changes the keyring constructor parameter's name from `type` to `defaultType`. This change makes it more clear that the keyring is not restricted to storing only one type of keys, but rather that new keys will be given this type if a type is not explicitly provided.

### DIFF
--- a/packages/keyring/src/index.spec.ts
+++ b/packages/keyring/src/index.spec.ts
@@ -20,7 +20,7 @@ describe('keypair', (): void => {
     let keypair: Keyring;
 
     beforeEach((): void => {
-      keypair = new Keyring({ type: 'ed25519' });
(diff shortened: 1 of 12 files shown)


jacogr:
General rule across all the polkadot-js repos: unless it is for a really good technical reason, we don't break userspace. As such, I won't be applying this. 

Especially in the case of the non-TypeScript users (where mismatches are not warned about), this would be very destructive since it will ignore what they specify and yield unexpected results all of a sudden. Unless there is -

- checking `type` as well, if found assigning to `defaultType`
- printing a warning in these cases that the option was renamed

Even with the above, the reason behind it is not convincing enough. If confusion, the docs can rather be updated instead of filtering this through the whole userbase without adding anything extra.

PR #496, opened 2019-12-12
Add cross client Encryption/Decryption

hskang9:
So there is an [EIP-1024](https://ethereum-magicians.org/t/eip-1024-cross-client-encrypt-decrypt/505) to provide private messaging in blockchain, but the idea has some vulnerability in that the encrypting keypair can be exposed between communication between client and Ethereum node. Eventually, I made cross-client encryption/decryption fully on the client using polkadot.js. This is an enhancement for privacy where clients can handle encrypted data between sender and receiver.


hskang9:
@jacogr please review

jacogr:
All on me - it was in my queue for yesterday, however npmjs mishaving killed my day (and made any publishes impossible). Sorry about that.

hskang9:
Whaaat I am trying to use it for encrypted private data exchange on blockchain though.

jacogr:
So not sure what your comment refers to here -

1. Yes, it was merged (so it is in a beta already)
2. Yes, we have exactly the boxing in initially - however, it was not used anywhere (so without real-world use, was removed)

hskang9:
Referring 2, me surprising that the feature was gone because it can be used for something

jacogr:
The issue is that all code needs to be maintained - so when I started, added everything that may be useful at some point. And that is always an issue since you cannot predict the future. 

So really happy having it in and used - then we know it works as expected and the corner cases is catered for. 

So I have a whole list of stuff that “may someday be useful”, but code has a cost, so it is not in there :)

jacogr:
This PR is actually an excellent example of not being able to predict the future 100%, even if just test-wise. When it was in (briefly), we didn’t have the concept of derived paths yet. In your case you made very sure all that works as expected. 

(Not a major impact on the nacl side, since it takes an secret, so probably a bad example anyway - nacl is quite small and to the point)

hskang9:
Thank you @jacogr I will make sure I keep up the good work.

PR #504, opened 2020-01-15
Avoid default import

Swader:
This is a blocker when integrating with some libraries. Notably, it breaks the address-encoder build for ENS: https://github.com/ensdomains/address-encoder/pull/23

Changing to this syntax fixes things.

### DIFF
--- a/packages/util-crypto/src/key/DeriveJunction.ts
+++ b/packages/util-crypto/src/key/DeriveJunction.ts
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import BN from 'bn.js';
+import * as BN from 'bn.js';
 import { bnToHex, compactAddLength, hexToU8a, isBn, isHex, isNumber, isString, stringToU8a } from '@polkadot/util';
 
 import blake2AsU8a from '../blake2/asU8a';


jacogr:
What exactly breaks? What is the issue?

This is used in this form everywhere in the polkadot-js repo, just in this repo there are 38 occurrences in 29 files.

As it stands, cannot see how a single file changes that. Sadly, it could also be one of those thing being changed back (aka, "it looks weird, revert" 3 months from now so would at the very least need some description of what is avoids.)

Swader:
Specifically, from the build error:

```bash
node_modules/@polkadot/util-crypto/key/DeriveJunction.d.ts:1:8 - error TS1259: Module '"/home/travis/build/ensdomains/address-encoder/node_modules/@types/bn.js/index"' can only be default-imported using the 'esModuleInterop' flag
1 import BN from 'bn.js';
         ~~
  node_modules/@types/bn.js/index.d.ts:586:1
    586 export = BN;
        ~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.
```

Using the flag is not an option because many modules are added to that library, each of which supports its own chain's crypto, and so modules don't really get along if interop is on. At least that's my impression, this might mean something else entirely.

Arachnid:
AIUI, the issue here is that `import foo from 'module'` in JS imports the default export from that module. For modules that don't have a default export, it constructs an object with all of that module's exports as properties.

TypeScript takes exception to this automatic behaviour, and wants an explict declaration. `import * as foo from 'module'` is explicitly asking for all the exports to be assigned to the object foo. It shouldn't break anything - and indeed, will avoid breakage if any of those modules suddenly adds a default export.

jacogr:
It def doesn't break anything, agree, and understand the imports. Even when `esModuleInterop` (or `allowSyntheticDefaultImports` is set, this syntax works 100%. And personally it is very important to make sure to support all other blockchain projects out there... 

My point is we are changing a single occurrence of many here, so it is bound to rear it's head again as usage expands. (And probably on other libs used as well as well, but those I'd rather not even think of). So worried somebody is going to make a mistake and break the world :)

At the very least (until the next PR has to be made for the same issue on another file), would suggest adding a single line comment to just say `// allow use without esModuleInterop` - so somebody doesn't break it.

Swader:
We have found an alternative workaround for the original issue, which makes this less urgent. Would you consider accepting a more comprehensive PR which removes defaults from all 33ish occurences?

jacogr:
Yes, would certainly and readily merge that.

Swader:
Totally forgot about this 🤦‍♂️

PR #530, opened 2020-02-14
feat(dependencies): remove moment.js

AndreasGassmann:
It looks like moment.js was only used in one place. So it seems like a huge overkill to include such a big package.

I removed the dependency and added a small function to output the same formatted date string.


jacogr:
Thanks a million for this. I would suggest just adding a test to ensure that we are indeed getting the right output from the function - fat fingers in the future may break it.

AndreasGassmann:
I wanted to write tests for it, but then I saw that that there already is a regex that checks the date in every test for the logger.

`dateMatch = expect.stringMatching(/20[0-9]{2}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}/);`

So it doesn't test that function explicitly, but if it is changed and then returns the wrong result, the tests will still fail.

Is that ok or should I add some specific tests for it?

AndreasGassmann:
Looking at it again, I'm not sure how to do this best. My function is not exported, so I can't test it directly. So either I have to put it into a separate file or compare the log string with the current time. I'm just afraid that in the second case the tests might fail randomly if the 2 `new Date()` calls return a date with a different second.

jacogr:
The export would be fine for tests. (Or we can move it to top-level format as well). For the specific test, would just suggest passing a `new Date(123456)` (just something known)

Sorry for being a pain here, always easier to get the tests sooner rather than later (elsewhere it is sometimes an uphill battle, the common repo is relative decent in having proper coverage as a whole)

jacogr:
I am going to merge this as-is. Will make a small additional PR to -

- move the date formatting function to format/
- just add a test around it

(Want to get this in the 2.3 release today since it is a nice win)

PR #563, opened 2020-03-10
fix: BalanceFormatter signature

hanwencheng:
signature fix.

### DIFF
--- a/packages/util/src/format/formatBalance.ts
+++ b/packages/util/src/format/formatBalance.ts
@@ -26,7 +26,7 @@ interface Options {
 }
 
 interface BalanceFormatter {
-  <ExtToBn extends ToBn> (input?: number | string | BN | BigInt | ExtToBn, options?: Options, decimals?: number): string;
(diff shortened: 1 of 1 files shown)


packages/util/src/format/formatBalance.ts
```diff
@@ -26,7 +26,7 @@ interface Options {
 }
 
 interface BalanceFormatter {
-  <ExtToBn extends ToBn> (input?: number | string | BN | BigInt | ExtToBn, options?: Options, decimals?: number): string;
+  <ExtToBn extends ToBn> (input?: number | string | BN | BigInt | ExtToBn, options?: Options | boolean, decimals?: number): string;
```
jacogr: This is actually done on purpose - the old-style `boolean` is something that is just kept for historical purposes to not break userspace. (It is deprecated, not yet removed).

jacogr:
Not a bug, a feature. `Option` should be used in all code, `boolean` is historic & deprecated. Closing.

PR #578, opened 2020-04-01
Throws an error when signing with a locked keypair

### LINKED ISSUE #576: Throw an error when the keypair is locked
Currently the "locked" keypair has a secret key of zero:

https://github.com/polkadot-js/common/blob/697a452ba6be77b747dbf745bef790313ded8285/packages/keyring/src/pair/index.ts#L136-L138

`sign()` will return an invalid signature if a developer forgets to check `.isLocked` before signing:

https://github.com/polkadot-js/common/blob/697a452ba6be77b747dbf745bef790313ded8285/packages/keyring/src/pair/index.ts#L142-L143

 This is confusing to new developers. It's better to raise an error instead of returning a wrong signature here.

h4x3rotab:
Fixes #576

### DIFF
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -62,6 +62,10 @@ function verify (type: KeypairType, message: Uint8Array, signature: Uint8Array, 
     : naclVerify(message, signature, publicKey);
 }
 
+function isLocked (secretKey: Uint8Array): boolean {
+  return !secretKey || secretKey.length === 0 || isEmpty(secretKey);
+}
+
 /**
  * @name pair
  * @summary Creates a keyring pair object
@@ -102,7 +106,7 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
       return meta;
     },
     get isLocked (): boolean {
-      return !secretKey || secretKey.length === 0 || isEmpty(secretKey);
+      return isLocked(secretKey);
     },
     get publicKey (): Uint8Array {
       return publicKey;
@@ -124,8 +128,9 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
       }
     },
     derive: (suri: string, meta?: KeyringPair$Meta): KeyringPair => {
-      assert(secretKey, 'Cannot derive on a locked keypair');
-
+      if (isLocked(secretKey)) {
+        throw new Error('Cannot derive on a locked keypair');
+      }
       const { path } = keyExtractPath(suri);
       const derived = keyFromPath({ publicKey, secretKey: secretKey }, path, type);
 
@@ -139,8 +144,12 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
     setMeta: (additional: KeyringPair$Meta): void => {
       meta = { ...meta, ...additional };
     },
-    sign: (message: Uint8Array, options?: SignOptions): Uint8Array =>
-      sign(type, message, { publicKey, secretKey }, options),
+    sign: (message: Uint8Array, options?: SignOptions): Uint8Array => {
+      if (isLocked(secretKey)) {
+        throw new Error('Cannot sign with a locked key pair');
+      }
+      return sign(type, message, { publicKey, secretKey }, options)
+    },
     toJson: (passphrase?: string): KeyringPair$Json =>
       toJson(type, { meta, publicKey }, encode({ publicKey, secretKey }, passphrase), !!passphrase),
     verify: (message: Uint8Array, signature: Uint8Array): boolean =>
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -6,7 +6,7 @@ import { Keypair, KeypairType } from '@polkadot/util-crypto/types';
 import { KeyringPair, KeyringPair$Json, KeyringPair$Meta, SignOptions } from '../types';
 import { PairInfo } from './types';
 
-import { u8aConcat, assert } from '@polkadot/util';
+import { u8aConcat } from '@polkadot/util';
 import { keyExtractPath, keyFromPath, naclKeypairFromSeed as naclFromSeed, naclSign, naclVerify, schnorrkelKeypairFromSeed as schnorrkelFromSeed, schnorrkelSign, schnorrkelVerify } from '@polkadot/util-crypto';
 
 import decode from './decode';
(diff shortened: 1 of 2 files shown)


packages/keyring/src/pair/index.ts
```diff
@@ -124,8 +128,9 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
       }
     },
     derive: (suri: string, meta?: KeyringPair$Meta): KeyringPair => {
-      assert(secretKey, 'Cannot derive on a locked keypair');
-
+      if (!secretKey || isLocked(secretKey)) {
```
jacogr, line 131:
```suggestion
      if (isLocked(secretKey)) {
```

`isLocked` already checks for `!secretKey`

packages/keyring/src/pair/index.ts
```diff
@@ -62,6 +62,10 @@ function verify (type: KeypairType, message: Uint8Array, signature: Uint8Array,
     : naclVerify(message, signature, publicKey);
 }
 
+function isLocked (secretKey?: Uint8Array): boolean {
```
jacogr, line 65:
```suggestion
function isLocked (secretKey?: Uint8Array): secretKey is Uint8Array {
```

If passing, it will be seen as a valid u8a

packages/keyring/src/pair/index.ts
```diff
@@ -139,8 +144,12 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
     setMeta: (additional: KeyringPair$Meta): void => {
       meta = { ...meta, ...additional };
     },
-    sign: (message: Uint8Array, options?: SignOptions): Uint8Array =>
-      sign(type, message, { publicKey, secretKey }, options),
+    sign: (message: Uint8Array, options?: SignOptions): Uint8Array => {
+      if (isLocked(secretKey)) {
+        throw new Error('Cannot sign with a locked key pair');
+      }
```
jacogr, line 150:
```suggestion
      assert(!isLocked(secretKey), 'Cannot sign with a locked key pair');
      
```

packages/keyring/src/pair/index.ts
```diff
@@ -124,8 +128,9 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
       }
     },
     derive: (suri: string, meta?: KeyringPair$Meta): KeyringPair => {
-      assert(secretKey, 'Cannot derive on a locked keypair');
-
+      if (!secretKey || isLocked(secretKey)) {
+        throw new Error('Cannot derive on a locked keypair');
+      }
```
jacogr, line 133:
```suggestion
      assert(!isLocked(secretKey), 'Cannot derive on a locked keypair');
      
```

isLocked already checks for !secretKey

packages/keyring/src/pair/index.ts
```diff
@@ -6,7 +6,7 @@ import { Keypair, KeypairType } from '@polkadot/util-crypto/types';
 import { KeyringPair, KeyringPair$Json, KeyringPair$Meta, SignOptions } from '../types';
 import { PairInfo } from './types';
 
-import { u8aConcat, assert } from '@polkadot/util';
+import { u8aConcat } from '@polkadot/util';
```
jacogr, line 9:
```suggestion
import { assert, u8aConcat } from '@polkadot/util';
```

Includes `assert` used below.

packages/keyring/src/pair/index.ts
h4x3rotab: Done. Good suggestion

packages/keyring/src/pair/index.ts
h4x3rotab: Done

packages/keyring/src/pair/index.ts
h4x3rotab: Done

packages/keyring/src/pair/index.ts
h4x3rotab: Done

packages/keyring/src/pair/index.ts
```diff
@@ -124,10 +128,10 @@ export default function createPair ({ toSS58, type }: Setup, { publicKey, secret
       }
     },
     derive: (suri: string, meta?: KeyringPair$Meta): KeyringPair => {
-      assert(secretKey, 'Cannot derive on a locked keypair');
+      assert(!isLocked(secretKey), 'Cannot derive on a locked keypair');
 
       const { path } = keyExtractPath(suri);
-      const derived = keyFromPath({ publicKey, secretKey: secretKey }, path, type);
+      const derived = keyFromPath({ publicKey, secretKey: secretKey! }, path, type);
```
jacogr, line 134:
```suggestion
      const derived = keyFromPath({ publicKey, secretKey: secretKey }, path, type);
```

jacogr: Sorry about this - the linter does complain, but since the isLocked now does it properly, we are ok and can cleanup.

h4x3rotab: I hope I would know more about Typescript but the linter is still complaining. Let me do some homework :)

jacogr: Grrrr... so that just says the value should not be empty. And the assert/isLocked checks for that - so should just work as expected.

h4x3rotab: The linter doesn't allow `secretKey` without the not-null assertion.

Just read the doc about [type predicates](https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates). It looks like a function with a type predicate return type notation is used for narrowing down a type. And the predicates should be used in compiling time. However `isLocked` is used in runtime because it also checks the content of `secretKey`. So it sounds to me a better way is to return a boolean and add change it back to `secretKey: secretKey!`.

jacogr: Something is funny here. But anyway, a warning (although it is only that way since al the `!`'s have not been cleaned - so many hours in a day).

Not going to waste your time, going to get it is since it works.

jacogr: *slush* Was my bug, not yours.

So `isLocked(Uint8Array | undefined)` is true if not a valid (for us), Uint8Array. So effectively we should return "it is not a (valid) Uint8Array".

Adjusting it somewhat in my linting branch.

PR #587, opened 2020-04-24
Create createKeyMulti & createKeySub for utility (util-crypto)

jacogr:
cc @danforbes


danforbes:
Now that's what I call service...thanks, man!

Issue #506, opened 2020-01-23
Support ECDSA(secp256k1) keypairs


akru:
**Motivation**

Currently Polkadot/Substrate [support](https://github.com/paritytech/substrate/blob/bc85d352681018e5b19369b758c013a52c5c3032/primitives/runtime/src/lib.rs#L166) three kinds of cryptographic primitives as `MultiSignature` data type:

* Ed25519
* Sr25519
* ECDSA

Unfortunately, polkadot-js support [first two](https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/types.ts) only. It's limiting using ECDSA keys for `subkey` CLI utility, which makes user experience pipeline a bit difficult.

**Proposal**

Introduce ECDSA keypair into polkadot-js crypto modules to support `secp256k1` crypto keys as same as Ed/Sr25519.

jacogr:
Thanks for logging, it has been in the back on my mind, not sure why it was not logged. 

PRs are very welcome on this one, would love support, but it probably won’t happen soon.

lovesh:
I would like to contribute this and submit over the next week. Anyone else already working on this?

akru:
@lovesh I’m working on this issue as part of ECDSA integration grant.

lovesh:
@akru Thanks for this.

jacogr:
Closed in #589, released in 2.9.1

PR #589, opened 2020-04-26
Added ECDSA keypair support

akru:
**Motivation**

Support ECDSA Substrate signatures according to https://github.com/polkadot-js/common/issues/506. 

**Changes**

* Added sign/verify functions into `util-crypto`.
* Added `ecdsa` keypair support into `keyring`.
* Added tests.


akru:
@jacogr Got it, I'll fix lint soon.

packages/keyring/src/keyring.ts
```diff
@@ -40,7 +40,7 @@ export default class Keyring implements KeyringInstance {
   constructor (options: KeyringOptions = {}) {
     options.type = options.type || 'ed25519';
 
-    assert(options && ['ed25519', 'sr25519'].includes(options.type || 'undefined'), `Expected a keyring type of either 'ed25519' or 'sr25519', found '${options.type}`);
+    assert(options && ['ed25519', 'sr25519', 'ecdsa'].includes(options.type || 'undefined'), `Expected a keyring type of either 'ed25519', 'sr25519' or 'ecdsa', found '${options.type}`);
```
jacogr, line 43:
```suggestion
    assert(options && ['ecdsa', 'ed25519', 'sr25519'].includes(options.type || 'undefined'), `Expected a keyring type of either 'ed25519', 'sr25519' or 'ecdsa', found '${options.type}`);
```

Text is fine, however the check array just alphabetical.... hopefully we don't end up here with 100's, but you never know :)

packages/keyring/src/pair/index.ts
```diff
@@ -98,10 +112,14 @@ function isLocked (secretKey?: Uint8Array): secretKey is undefined {
  * an `encoded` property that is assigned with the encoded public key in hex format, and an `encoding`
  * property that indicates whether the public key value of the `encoded` property is encoded or not.
  */
-export default function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: PairInfo, meta: KeyringPair$Meta = {}, encoded: Uint8Array | null = null): KeyringPair {
+export default function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: PairInfo, meta: KeyringPair$Meta = {}, encoded: Uint8Array | null = null, address: Uint8Array | null = null): KeyringPair {
```
jacogr, line 115: Not convinced about this addition - since you can take the `publicKey` -> `address` (as per the helper above), why pass it explicitly? It has hoops elsewhere as well, i.e. we are actually passing an empty u8a for ecdsa in the construction.

packages/keyring/src/pairs.ts
```diff
@@ -13,7 +13,7 @@ export default class Pairs implements KeyringPairs {
   readonly #map: KeyringPairMap = {};
 
   public add (pair: KeyringPair): KeyringPair {
-    this.#map[pair.publicKey.toString()] = pair;
+    this.#map[decodeAddress(pair.address).toString()] = pair;
```
jacogr, line 16: We want to lookup on the underlying `publicKey` bytes - this allows addresses for any encoding to be retrieved. So you can add ssFormat 42 addresses, change to 1 and still retrieve and have it available.

packages/keyring/src/keyring.ts
```diff
@@ -85,9 +85,11 @@ export default class Keyring implements KeyringInstance {
    * `addPair` to stores in a keyring pair dictionary the public key of the generated pair as a key and the pair as the associated value.
    */
   public addFromAddress (address: string | Uint8Array, meta: KeyringPair$Meta = {}, encoded: Uint8Array | null = null, type: KeypairType = this.type, ignoreChecksum?: boolean): KeyringPair {
-    const publicKey = this.decodeAddress(address, ignoreChecksum);
+    address = this.decodeAddress(address, ignoreChecksum);
+    const publicKey = type === 'ecdsa'
+                    ? new Uint8Array() : address;
```
jacogr, line 90: Always pass the publicKey, don't handle it differently. Since here you are using the address as the public identifier anyway, pass it as such.

jacogr:
@akru Any progress on this front? Would love to see this in since it is a great effort.

akru:
Hey @jacogr !

Thank you for your review!

I've fix places with your comments around address: it was a problem because ECDSA public isn't equal to address when read key from json, but I make workaround here: https://github.com/polkadot-js/common/pull/589/files#diff-2e199d4d80314053b3cb0c4e1d764fb3R72, probably it will works.

jacogr:
Thanks a million, will take a peek fist thing in the morning .

Yes, I had some thoughts about my comments, it is a tough one either way, I understand why it went that way,

Will take a peek, ASAP and let's get it in :)

jacogr:
My humble apologies, despite best-effort I not manage to give this the attention it deserves today. Have this open for first thing in the morning.

PR #590, opened 2020-04-26
Add example for multisig address creation

joepetrowski:
Adds an example of how to use `createKeyMulti`.

### DIFF
--- a/docs/examples/util-crypto/04_generate_multisig_address/index.js
+++ b/docs/examples/util-crypto/04_generate_multisig_address/index.js
@@ -1,28 +1,29 @@
+/* eslint-disable header/header */
 import { createKeyMulti, encodeAddress } from '@polkadot/util-crypto';
 
-function main() {
-	// Input the addresses that will make up the multisig account. This utility
-	// will sort them for you but when you submit to a Substrate node, you must
-	// sort them yourself.
-	const addresses = [
-		'1nUC7afqmo7zwRFWxDjrUQu9skk6fk99pafb4SiyGSRc8z3',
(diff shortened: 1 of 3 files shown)


docs/examples/util-crypto/04_generate_multisig_address/index.js
```diff
@@ -0,0 +1,29 @@
+/* eslint-disable header/header */
+import { createKeyMulti, encodeAddress } from '@polkadot/util-crypto';
+
+function main () {
+  // Input the addresses that will make up the multisig account. This utility
```
jacogr, line 5: https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/address/sort.ts

jacogr: So with the above the comment is not "100%" correct, or rather, can show the used "how to sort" - since it needs to sort by the publicKey, not via ss58.

docs/examples/util-crypto/04_generate_multisig_address/index.js
```diff
@@ -1,27 +1,41 @@
...
+  // Convert byte array to SS58 encoding.
+  const Ss58Address = encodeAddress(multiAddress, SS58Prefix);
+
+  console.log(`\nMultisig Address: ${Ss58Address}`);
+
+  // Take addresses and remove the sender.
+  const otherSignatories = addresses.filter(who => who !== addresses[index]);
+
+  // Sort them by public key.
+  const otherSignatoriesSorted = sortAddresses(otherSignatories).map(
```
jacogr: Can just pass the prefix here, it is available in master (beta releases)

PR #610, opened 2020-06-04
Add BN consts and timeToString

kwingram25:


packages/util/src/bn/index.ts
```diff
@@ -6,6 +6,7 @@
  * @summary Utility methods to convert to and from `BN` objects
  */
 
+export * from './consts';
```
jacogr, line 9: Not sure about this since it is not user the BN_* namespace. So if the consts were `BN_<SOMETHING>` it would work.

packages/util/src/types.ts
```diff
@@ -40,3 +40,5 @@ export type BnList = {
   0: BN;
   1: BN;
 } & BN[];
+
+export type Time = [number, number, number, number];
```
jacogr, line 44: Maybe be explicit in a struct rather than a tuple? Can just see myself having to lookup.

jacogr:
Generally would really have liked tests here, the util is very high in coverage and it certainly makes it easier to maintain to just add them as they come in.

packages/util/src/extractTime.ts
```diff
@@ -0,0 +1,64 @@
...
+function addTime (a: Time, b: Time): Time {
+  return {
+    days: a.days + b.days,
+    hours: a.hours + b.hours,
+    minutes: a.minutes + b.minutes,
+    seconds: a.seconds + b.seconds
+  };
+}
+
+const ZERO = { days: 0, hours: 0, minutes: 0, seconds: 0 };
```
jacogr, line 24: Re-looking at this - this is actually quite "weird" since it doesn't have ms. JS dates are to the ms. (I understand this solves a specific problem, however we need to align - if the caller doesn't have ms, he needs to add them)

PR #619, opened 2020-06-19
Allow encoding with 0 as ss58 prefix

Tbaut:
The default value for this param is applied even if 0 (a valid argument) is passed as ss58Prefix.

### DIFF
--- a/packages/util-crypto/src/address/encode.spec.ts
+++ b/packages/util-crypto/src/address/encode.spec.ts
@@ -21,6 +21,12 @@ describe('encode', (): void => {
     ).toEqual('7sL6eNJj5ZGV5cn3hhV2deRUsivXfBfMH76wCALCqWj1EKzv');
   });
 
+  it('can re-encode an address to Polkadot live', (): void => {
+    expect(
+      encode('5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY', 0)
+    ).toEqual('15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5');
+  });
+
   it('fails when non-valid publicKey provided', (): void => {
     expect(
       (): string => encode(
--- a/packages/util-crypto/src/address/encode.ts
+++ b/packages/util-crypto/src/address/encode.ts
@@ -13,7 +13,9 @@ import decode from './decode';
 import defaults from './defaults';
 import sshash from './sshash';
 
-export default function encode (_key: Uint8Array | string, ss58Format: Prefix = defaults.prefix): string {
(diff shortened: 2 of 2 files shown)


jacogr:
Couple of issues here -

- you have broken the public-facing API
- the default is not applied on 0 (only undefined)
- just adding the test like you have (See https://github.com/polkadot-js/common/pull/620) passes, eg.

```
$ yarn test:one packages/util-crypto/src/address/encode.
$ polkadot-dev-run-test packages/util-crypto/src/address/encode.
 PASS  packages/util-crypto/src/address/encode.spec.ts
  encode
    ✓ encodes an address to a valid value (2 ms)
    ✓ can re-encode an address
    ✓ can re-encode an address to Polkadot live (1 ms)
    ✓ fails when non-valid publicKey provided (2 ms)
    ✓ encodes a 1-byte address
    ✓ encodes a 1-byte address (with prefix) (1 ms)
    ✓ encodes a 2-byte address
    ✓ encodes a 4-byte address
    ✓ enodes a 8-byte address

Test Suites: 1 passed, 1 total
Tests:       9 passed, 9 total
Snapshots:   0 total
Time:        3.716 s
Ran all test suites matching /packages\/util-crypto\/src\/address\/encode./i.
```

Tbaut:
hmm CI is complaining:
```bash
Using TypeScript 3.9.5 from /home/runner/work/common/common/node_modules/typescript/lib
Error: /home/runner/work/common/common/packages/keyring/src/keyring.ts(194)
 Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.
Error: /home/runner/work/common/common/packages/keyring/src/pair/nobody.ts(9)
 Expected 2 arguments, but got 1.
Error: /home/runner/work/common/common/packages/util-crypto/src/address/sort.ts(14)
 Argument of type 'number | undefined' is not assignable to parameter of type 'number'.
  Type 'undefined' is not assignable to type 'number'.
##[error]Process completed with exit code 255.
```

jacogr:
CI is complaining since you turned an optional value into a required value - see my first comment about breaking the public API.

Tbaut:
> CI is complaining since you turned an optional value into a required value - see my first comment about breaking the public API.

Makes sense.

> just adding the test like you have (See #620) with no changes passes, eg.
Encode is totally fine indeed, I finally found my original error.
I had problems with `encodeAddress`, closing.

jacogr:
There was an issue on `keyring.encodeAddress` where the ss58Format param is not passed as an option - https://github.com/polkadot-js/common/pull/620/files#diff-e0722dd4ea9a8954b71f51883937d8b3R195

PR #637, opened 2020-07-20
Keyring JSON with kdf

jacogr:
- [x] pbdkf2 for JS/WASM as available
- [x] verification that the flow is sane
- [x] update tests for new
- [x] include tests for old
- [x] include tests for conversion
- [x] end-to-end testing against real
- [x] end-to-end verification on upgrade
- [x] version bump to v3

Still TODO in follow-up PRs in the 3 series -

- WASM scrypt (https://docs.rs/scrypt/0.3.1/scrypt/), currently a bit problematic since the WASM build is still restructured and not quite stable with the upgrades to wasmpack
- https://github.com/polkadot-js/common/issues/638

### DIFF
--- a/packages/util-crypto/src/index.ts
+++ b/packages/util-crypto/src/index.ts
@@ -20,6 +20,7 @@ export * from './nacl';
 export * from './pbkdf2';
 export * from './random';
 export * from './schnorrkel';
+export * from './scrypt';
 export * from './secp256k1';
 export * from './sha512';
 export * from './signature';
--- a/packages/util-crypto/src/index.ts
+++ b/packages/util-crypto/src/index.ts
@@ -17,6 +17,7 @@ export * from './keccak';
 export * from './key';
 export * from './mnemonic';
 export * from './nacl';
+export * from './pbkdf2';
 export * from './random';
 export * from './schnorrkel';
 export * from './secp256k1';
--- a/packages/util-crypto/src/scrypt/fromU8a.ts
+++ b/packages/util-crypto/src/scrypt/fromU8a.ts
@@ -4,7 +4,9 @@
 
 import { Params } from './types';
 
-import { u8aToBn } from '@polkadot/util';
+import { u8aToBn, assert } from '@polkadot/util';
+
+import { DEFAULT_PARAMS } from './defaults';
 
 interface Result {
   params: Params,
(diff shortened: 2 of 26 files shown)


packages/util-crypto/src/scrypt/fromU8a.ts
```diff
@@ -0,0 +1,21 @@
...
+import { u8aToBn } from '@polkadot/util';
+
+interface Result {
+  params: Params,
+  salt: Uint8Array;
+}
+
+export default function scryptFromU8a (data: Uint8Array): Result {
+  const salt = data.subarray(0, 32);
+  const N = u8aToBn(data.subarray(32 + 0, 32 + 4), { isLe: true }).toNumber();
```
kirushik, line 16: I wonder if it would make sense to add some sanity checks here to avoid DoSing by some ridiculously-high numbers.
Probably is not an issue in client-side usage, though, since you can always close the tab.
But if we ever use this code serverside, maybe it would make sense to either explicitly timeout the derivation function or put some reasonable limits (like all defaults + 2 maybe?) on the values here?

packages/util-crypto/src/index.ts
```diff
@@ -17,8 +17,10 @@ export * from './keccak';
 export * from './key';
 export * from './mnemonic';
 export * from './nacl';
+export * from './pbkdf2';
```
kirushik, line 20: Do you still want to keep it, even despite PBKDF2 is not being used anymore?

jacogr: Yes, since it is actually used elsewhere it should have been exposed already and it wasn't.

packages/util-crypto/src/scrypt/fromU8a.ts
jacogr: Yes, this is not great an opens up a hole. For now, I'm going to fix it to the defaults until there is a proper plan of attack - need to really look around on best-practices for this. User inputs are always an issue...

```
// FIXME At this moment we assume these to be fixed params, this is not a great idea since we lose flexibility
  // and updates for greater security. However we need some protection against carefully-crafted params that can
  // eat up CPU since these are user inputs. So we need to get very clever here, but atm we only allow the defaults
  // and if no match, bail out
  assert(N === DEFAULT_PARAMS.N && p === DEFAULT_PARAMS.p && r === DEFAULT_PARAMS.r, 'Invalid injected scrypt params found');
```

PR #641, opened 2020-07-24
Change formatBalance s.t. Unit is prefix instead of postfix

retotrinkler:
**Purely subjective PR**

As far as I can tell the [international number format](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) has the symbol as prefix instead of postfix. They also mention something in [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217#Position_of_ISO_4217_code_in_amounts) on this but also state that it depends on country. Think it would be nice to go with English international convention and thought would create this PR.

```js
console.log(new Intl.NumberFormat('en-IN', { style: 'currency', currency: 'EUR' }).format(number));
console.log(new Intl.NumberFormat('en-IN', { style: 'currency', currency: 'USD' }).format(number));
console.log(new Intl.NumberFormat('en-IN', { style: 'currency', currency: 'CHF' }).format(number));

// Outputs: 
// "€1,23,456.79"
// "US$1,23,456.79"
// "CHF 1,23,456.79"
```

I tried to adjust the specs but did not run any tests.

### DIFF
--- a/packages/util/src/format/formatBalance.spec.ts
+++ b/packages/util/src/format/formatBalance.spec.ts
@@ -18,25 +18,25 @@ describe('formatBalance', (): void => {
     it('formats 123,456,789,000 (decimals=15)', (): void => {
       expect(
         formatBalance(TESTVAL, { decimals: 15, withSi: true })
-      ).toEqual('123.456µ Unit');
+      ).toEqual('Unit 123.456µ');
     });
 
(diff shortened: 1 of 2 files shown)


jacogr:
Closing this one. Unlikely to go it at all. I understand the why, however the format is pretty much fixed at this point at lest for the libraries here.

Issue #656, opened 2020-08-10
Support for Ethereum compatible ECDSA 


crystalin:
Current ECDSA configuration in the PolkadotJS is using Blake2 Hash into H256 (which is also the setup for Substrate default ECDSA)

In order to support Ethereum compatible chains (Like https://github.com/paritytech/frontier pallet offers), it needs to have a Eth compatible implementation of the ECDSA using Keccak256 into H160 Account ID.
This would also require to skip the base58 encoding (as Eth addresses are not commonly using base58)

I think in order to maintain current ECDSA, it would be good to include a forth type: `ecdsa-eth` (others being "sr25519", "ed25519", "ecdsa"), but I'm open to other solutions.

Most of the changes to support it needs to be done in the @polkadot/util-crypto/address and @polkadot/keyring

jacogr:
ECDSA support was added in https://github.com/polkadot-js/common/pull/589 - the reason for linking that is that the support would be exactly the same in terms or areas touched and tests added. (Also a good example since the first two were done over time and alongside each-other) 

Here there are some shortcuts which can be taken, since this deals with mostly hashing changes a lot of the functions (added above) can just be extended with an additional param that specified the hash type, for example 

`secp256k1Sign (message: Uint8Array | string, { secretKey }: Partial<Keypair>)`

becomes

`secp256k1Sign (message: Uint8Array | string, { secretKey }: Partial<Keypair>, hasher = 'blake2' | 'keccak')`

(Alongside adjustment of tests and the addition of new)

Overall def. something to consider as a new `ethereum` keypair.

crystalin:
Thank you @jacogr , that sounds good.
Is it something your team has time to work on ?
I can probably try to come with a PR but not being familiar with your code, it will take me more time.

jacogr:
PRs are always preferred and welcome. Since the polkadot-js team is effectively me on all the repos hosted here, I do rely on contributions for non-as-shipped-core a lot. I do get to everything, but no promises as to "when".

crystalin:
Oh, you have done a lot of work on it already. Thanks
I believe this will be directly usable once the #661 is merged as it includes the "ethereum" type.
Let me know if you want me to test the branch

crystalin:
Good work on those PR Jaco !
I've spotted another place where the address doesn't get converted correctly:
[...]

PR #661, opened 2020-08-12
Ethereum keypair support

jacogr:
Part of https://github.com/polkadot-js/common/issues/656

Sadly this is where things starts going off-the-rails -

- MultiSignature doesn't have a specific Ethereum byte (Potentially same as normal ecdsa?)
- Not sure about the publicKey hashing
- the actual generated pairs from the same seed is non-Ethereum compatible

### DIFF
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -7,7 +7,7 @@ import { KeyringPair, KeyringPair$Json, KeyringPair$JsonEncodingTypes, KeyringPa
 import { PairInfo } from './types';
 
 import { assert, u8aConcat } from '@polkadot/util';
-import { keyExtractPath, keyFromPath, naclKeypairFromSeed as naclFromSeed, naclSign, naclVerify, schnorrkelKeypairFromSeed as schnorrkelFromSeed, schnorrkelSign, schnorrkelVerify, secp256k1KeypairFromSeed as secp256k1FromSeed, secp256k1Sign, secp256k1Verify, blake2AsU8a } from '@polkadot/util-crypto';
+import { blake2AsU8a, ethereumEncode, keccakAsU8a, keyExtractPath, keyFromPath, naclKeypairFromSeed as naclFromSeed, naclSign, naclVerify, schnorrkelKeypairFromSeed as schnorrkelFromSeed, schnorrkelSign, schnorrkelVerify, secp256k1KeypairFromSeed as secp256k1FromSeed, secp256k1Sign, secp256k1Verify } from '@polkadot/util-crypto';
 
 import decode from './decode';
 import encode from './encode';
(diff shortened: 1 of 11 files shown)


packages/keyring/src/pair/index.ts
```diff
@@ -23,6 +23,9 @@ const SIG_TYPE_ED25519 = new Uint8Array([0]);
 const SIG_TYPE_SR25519 = new Uint8Array([1]);
 const SIG_TYPE_ECDSA = new Uint8Array([2]);
 
+// FIXME This needs a mapping to supported - maybe ecdsa for MultiSignature?
+const SIG_TYPE_ETHEREUM = new Uint8Array([255]);
```
jacogr, line 27:
```suggestion
const SIG_TYPE_ETHEREUM = SIG_TYPE_ECDSA;
```

jacogr:
Ok, this is now the second time that specific test failed. Need to compare against original and see what is weird - something changed that makes it behave erratically now.

jacogr:
Merging as-is. It does address some latent issues and doesn't break existing (although the `ethereum` type most probably still has holes that can only be fixed by testing and use)

crystalin:
@jacogr in keyring.ts there is a check in the constructor, it is missing the "ethereum" in the list

crystalin:
@jacogr Also the ethereum type gets converted to base58 which I think should be avoided

crystalin:
Last part, it also fails in the isEthereumAddress check, because the address is 66 bytes long. Ex: `0x11539e59d378f29b4cc073e85d4855d32a98df0f1e05c1887e77c5027d47718f`

crystalin:
@jacogr Also the publickey passed to the pair is only the first part of the public key (Ex: `0x03624f720eae676a04111631c9ca338c11d0f5a80ee42210c6be72983ceb620fbf`).
I believe the keccak requires the fullkey (0x624f720eae676a04111631c9ca338c11d0f5a80ee42210c6be72983ceb620fbf645a96f951529fa2d70750432d11b7caba5270c4d677255be90b3871c8c58069) to retrieve the address

jacogr:
1. Yes, it is known since it doesn't do Ethereum-compatible (follow-up PR does add basic support, ex derivation)
2. Clueless as to what you mean - address has Ethereum encoding (tests added in a follow-up)
3. publicKey !== address - so that function is for checking H160 addresses, not public keys (same with the encoding here, it only does address -> address encoding, follow-up does address and/or publicKey for at least address encoding since it became needed)
4. Indeed. This PR does not have expand/compress functions.

On the last point: most of the exposed keypair functionality is untested and stubbed with TODOs, until the tests are there, don't expect miracles, I'll get to it as I have time and expand. As examples: verify/sign here and in the follow-up is still untested and seemingly broken as well, there is still a number of PRs and tests left, saving is also problematic atm, or at least recovering from JSON)

crystalin:
Thank you @jacogr 
1. Great :)
2. Sorry this one is fine, I thought the address once sent gets converted to base58 but I was wrong.
3. Yes I know, I'm just stating what I observe when trying to use the ethereum type. Somewhere it either sends the wrong variable or uses the wrong test.
4. 👍 

Yes I know, I try to provide early feedback so it easier to move forward :)

PR #664, opened 2020-08-17
Ethereum keypair signing & verification

jacogr:
Part of https://github.com/polkadot-js/common/issues/656

Signing/Verification & JSON decoding tests for Ethereum pairs.


packages/keyring/src/pair/index.ts
```diff
@@ -48,7 +48,7 @@ const TYPE_VERIFY = {
     secp256k1Verify(message, signature, blake2AsU8a(publicKey), 'blake2'),
   ed25519: naclVerify,
   ethereum: (message: Uint8Array, signature: Uint8Array, publicKey: Uint8Array) =>
-    secp256k1Verify(message, signature, keccakAsU8a(publicKey), 'keccak'),
+    secp256k1Verify(message, signature, keccakAsU8a(secp256k1Compress(publicKey)), 'keccak'),
```
jacogr, line 51: Here - should be actual address. (And should apply to lower-levels as well)

PR #666, opened 2020-08-19
Fix `Error: @polkadot/wasm-crypto has not been initialized`

Bitcoinera:
Fixes this error, related to @wasm-crypto not being initialized:

image

### DIFF
--- a/packages/util-crypto/src/schnorrkel/deriveHard.ts
+++ b/packages/util-crypto/src/schnorrkel/deriveHard.ts
@@ -6,12 +6,13 @@ import { Keypair } from '../types';
 
 import '../polyfill';
 
-import { sr25519DeriveKeypairHard } from '@polkadot/wasm-crypto';
+import { sr25519DeriveKeypairHard, waitReady } from '@polkadot/wasm-crypto';
 
 import keypairFromU8a from './keypair/fromU8a';
 import keypairToU8a from './keypair/toU8a';
 
-export default function deriveHard (keypair: Keypair, chainCode: Uint8Array): Keypair {
+export default async function deriveHard (keypair: Keypair, chainCode: Uint8Array): Keypair {
+  await waitReady()
   return keypairFromU8a(
     sr25519DeriveKeypairHard(keypairToU8a(keypair), chainCode)
   );
--- a/packages/util-crypto/src/schnorrkel/derivePublic.ts
+++ b/packages/util-crypto/src/schnorrkel/derivePublic.ts
@@ -4,8 +4,9 @@
 
 import '../polyfill';
 
-import { sr25519DerivePublicSoft } from '@polkadot/wasm-crypto';
+import { sr25519DerivePublicSoft, waitReady } from '@polkadot/wasm-crypto';
 
-export default function deriveSoft (publicKey: Uint8Array, chainCode: Uint8Array): Uint8Array {
+export default async function deriveSoft (publicKey: Uint8Array, chainCode: Uint8Array): Uint8Array {
+  await waitReady()
   return sr25519DerivePublicSoft(publicKey, chainCode);
 }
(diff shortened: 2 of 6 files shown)


Bitcoinera:
Oops, the checks :eyes:

packages/util-crypto/src/schnorrkel/deriveHard.ts
```diff
@@ -6,12 +6,13 @@ import { Keypair } from '../types';
...
 
-import { sr25519DeriveKeypairHard } from '@polkadot/wasm-crypto';
+import { sr25519DeriveKeypairHard, waitReady } from '@polkadot/wasm-crypto';
 
 import keypairFromU8a from './keypair/fromU8a';
 import keypairToU8a from './keypair/toU8a';
 
-export default function deriveHard (keypair: Keypair, chainCode: Uint8Array): Keypair {
+export default async function deriveHard (keypair: Keypair, chainCode: Uint8Array): Keypair {
+  await waitReady()
```
jacogr, line 15: The issue with the fix above is that the signature is changed from sync to async. So the return here is not `Promise<Keypair>`

If you use sr25519, it is up to the user to initialise the wasm correctly with a wait. (e.g. via the API, the isReady does that, outside of that you need to do it manually)

jacogr:
This "error" is completely expected. It is up to the caller to initialize.

Bitcoinera:
> This "error" is completely expected. It is up to the caller to initialize.

Ok, so this is our code:

```
// creates a polkadot address
async function createPolkadotAddress(seedPhrase, network, curve) {
  const [{ Keyring }] = await Promise.all([
    import("@polkadot/api"),
    import("@polkadot/wasm-crypto").then(async ({ waitReady }) => {
      await waitReady()
    }),
    import("@polkadot/util-crypto").then(async ({ cryptoWaitReady }) => {
      // Wait for the promise to resolve, async WASM or `cryptoWaitReady().then(() => { ... })`
      await cryptoWaitReady()
    }),
  ])

  const keyring = new Keyring({
    ss58Format: Number(network.address_prefix),
    type: curve,
  })
  const newPair = keyring.addFromUri(seedPhrase)

  return {
    cosmosAddress: newPair.address,
    publicKey: newPair.publicKey,
    seedPhrase,
    curve,
  }
}
```

This is still failing, although we are waiting on initialization.

If I do this:

```
  try {
    const keyring = new Keyring({
      ss58Format: Number(network.address_prefix),
      type: curve,
    })
    const newPair = keyring.addFromUri(seedPhrase)
  } catch(error) {
    console.log('ERROR')
  }
```

I see the error comes from this Keyring creation. So I should convert this chunk into a Promise then or how should it work?

Thanks

jacogr:
wasm-crypto/waitReady and util-crypto/cryptoWaitReady are 100% equivalent, you don't need to use both.

Bitcoinera:
> wasm-crypto/waitReady and util-crypto/cryptoWaitReady are 100% equivalent, you don't need to use both.

Yes, but my point is that using those it is not working, the keyring part is still failing  :shrug: I think because of this import inside polkadot/util-crypto. 

Or maybe this chunk inside wasm-crypto should change?

```
  const wrapReady = (fn) =>
    (...params) => {
      assert(stubbed.isReady(), INIT_ERRROR);

      return fn(...params);
    };
```

jacogr:
The WASM async works as-is,  a good example where it is used without the API is eg. in the polkadot-js extension, there all pairs are sr25519.

So would suggest simplifying first and then working from there -

```js
import { Keyring } from '@polkadot/keyring';
import { cryptoWaitReady } from '@polkadot/util-crypto';

// creates a polkadot address
async function createPolkadotAddress(seedPhrase, network, curve) {
  await cryptoWaitReady();
  
const keyring = new Keyring({
    ss58Format: Number(network.address_prefix),
    type: curve,
  });

  const newPair = keyring.addFromUri(seedPhrase)
...
```

Make sure that you don't have multiple versions of the wams-crypto and keyring around. (It could be that you are inatilizing one, but one one that the api Keyring has)

Bitcoinera:
> So would suggest simplifying first and then working from there -

OK, that works. Maybe we weren't really waiting for `cryptoWaitReady` there. Your code does work though. Thank you so much Jaco.

Closing this here

PR #698, opened 2020-09-30
Adding usage example

roccomuso:


packages/util-crypto/README.md
```diff
@@ -16,6 +16,20 @@ Functions can be imported as follows:
...
 
-[Utilities](SUMMARY.md)
+```js
+import { mnemonicGenerate, mnemonicToBip39, naclKeypairFromSeed, encodeAddress } from '@polkadot/util-crypto';
+
+const mnemonic = mnemonicGenerate()
+console.log('mnemonic:', mnemonic)
+
+const seed = Buffer.from(mnemonicToBip39(mnemonic))
+console.log('seed', seed.toString('hex'))
```
jacogr, line 28:
```suggestion
import { u8aToHex } from '@polkadot/util';
import { mnemonicGenerate, mnemonicToBip39, naclKeypairFromSeed, encodeAddress } from '@polkadot/util-crypto';

const mnemonic = mnemonicGenerate()
console.log('mnemonic:', mnemonic)

const seed = u8aToHex(mnemonicToBip39(mnemonic))
console.log('seed', seed.toString('hex'))
```

packages/util-crypto/README.md
```diff
@@ -16,6 +16,20 @@ Functions can be imported as follows:
...
+## Example
 
-[Utilities](SUMMARY.md)
+```js
+import { mnemonicGenerate, mnemonicToBip39, naclKeypairFromSeed, encodeAddress } from '@polkadot/util-crypto';
+
+const mnemonic = mnemonicGenerate()
+console.log('mnemonic:', mnemonic)
+
+const seed = Buffer.from(mnemonicToBip39(mnemonic))
```
jacogr, line 27: We generally never want to use this one - it is only there for Ethereum-compatibility (i.e. like in Moonbeam). For Substrate/Polkadot chains this is never used. (This loses entropy along the way, hence it is done differently for Substrate)

packages/util-crypto/README.md
```diff
@@ -16,6 +16,20 @@ Functions can be imported as follows:
...
+console.log('seed', seed.toString('hex'))
+
+const keyPair = naclKeypairFromSeed(seed) // using ed25519
```
jacogr, line 30: Follow-up - since seed above is not Substrate-compatible, this would be different from anything else generated in the ecosystem, including anything using the JS libs such as the keyring, UIs, extensions.

jacogr (commented):
Thank you.

Couple of small comments.

packages/util-crypto/README.md
jacogr: See this for the seed via minisecret - https://github.com/polkadot-js/common/blob/master/packages/keyring/src/keyring.ts#L185-L187

packages/util-crypto/README.md
roccomuso: technically would work if I'd provide a minisecret seed instead, right?

packages/util-crypto/README.md
roccomuso: I see, so this is used as seed instead: https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/mnemonic/toMiniSecret.ts

roccomuso: I can't see substantial differences compared to https://github.com/bitcoinjs/bip39/blob/master/src/index.js#L55
Does it just include a default salt/password? https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/mnemonic/toMiniSecret.ts#L18

jacogr: The big difference is that the minisecret variant goes to entropy (which the older version doesn't) and then uses the rounds with the optional password. So in pure JS (assuming no WASM), -

- substrate version uses mnemonicToEntropy from bip39 and then applies hashing
- the Ethereum version uses mnemonicToSeed from bip39

roccomuso: is there any drawback in keeping using old seed generation?

jacogr: It is not compatible with anything in the ecosystem, none of the wallets (apps, Lunie, Enzyme, Math, ...), none of the tools (e.g. subkey) supports the legacy version. But apart from that, cannot comment on the implications, it is good for other chains, so shouldn't have a security mpact.

packages/util-crypto/README.md
jacogr: Yes, 100%. The mini-secret naming aligns just with the rest of substrate, but it, like the legacy format, is 32-bytes in length, so fits in perfectly.

roccomuso:
Another doubt I have regarding the util-crypto `encodeAddress(publicKey)` method. It does accept a prefix as second argument, by default it uses `0`. But I was expecting 42 (mainnet?).

But then I saw on https://polkadot.js.org/apps/#/accounts that they're using the 0 prefix to encode the address (is that a mainnet addr?)

jacogr:
0 = Polkadot, 2 = Kusama, the default is 42 which is a generic Substrate format not tied to a chain.

jacogr:
Closing. Docs move to docs repo (has examples for there and with the keyring).

PR #705, opened 2020-10-12
Add evmToAddress and addressToEvm crypto utils.

jnaviask:
Adds two utils for manipulating EVM addresses.

* addressToEvm is based on logic here: https://github.com/paritytech/substrate/blob/frontier/frame/evm/src/lib.rs#L148 and truncates a Substrate public key to a 20-byte Ethereum key.
* evmToAddress is based on logic here: https://github.com/paritytech/substrate/blob/frontier/frame/evm/src/lib.rs#L169 and hashes a 20-byte Ethereum key to produce a 32-byte Substrate public key.

### DIFF
--- a/packages/util-crypto/src/address/addressToEvm.ts
+++ b/packages/util-crypto/src/address/addressToEvm.ts
@@ -1,0 +1,14 @@
+// Copyright 2017-2020 @polkadot/util-crypto authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import decode from './decode';
+
+/**
+ * @name addressToEvm
+ * @summary Converts an SS58 address to its corresponding EVM address.
+ */
+export default function addressToEvm (address: string | Uint8Array, ignoreChecksum?: boolean): Uint8Array {
+  const decoded = decode(address, ignoreChecksum);
+
+  return decoded.subarray(0, 20);
+}
--- a/packages/util-crypto/src/address/evmToAddress.ts
+++ b/packages/util-crypto/src/address/evmToAddress.ts
@@ -1,0 +1,28 @@
+// Copyright 2017-2020 @polkadot/util-crypto authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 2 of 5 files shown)


packages/util-crypto/src/address/addressToEvm.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2017-2020 @polkadot/util-crypto authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import decode from './decode';
+
+/**
+ * @name addressToEvm
+ * @summary Converts an SS58 address to its corresponding EVM address.
+ */
+export default function (address: string | Uint8Array, ignoreChecksum?: boolean): Uint8Array {
```
jacogr, line 10:
```suggestion
export default function addressToEvm (address: string | Uint8Array, ignoreChecksum?: boolean): Uint8Array {
```

packages/util-crypto/src/address/evmToAddress.ts
```diff
@@ -0,0 +1,28 @@
...
+import hasher from '../secp256k1/hasher';
+import { HashType } from '../secp256k1/types';
+
+import { u8aConcat } from '@polkadot/util';
+
+/**
+ * @name evmToAddress
+ * @summary Converts an EVM address to its corresponding SS58 address.
+ */
+export default function (evmAddress: string | Uint8Array, ss58Format?: Prefix, hashType: HashType = 'blake2'): string {
```
jacogr, line 16:
```suggestion
export default function evmToAddress (evmAddress: string | Uint8Array, ss58Format?: Prefix, hashType: HashType = 'blake2'): string {
```

jacogr (commented):
Looks really good, thank you. Just made 2 mall comments exposing the name inside the files (these do appear in stack traces, depending on the import, so is generally useful).

jacogr (approved):
Thanks an absolute million. The people blazing the trail after you will have a much easier time because of this.

PR #763, opened 2020-11-10
Add eth tests and truncate address for keccak

joelamouche:
This PR will fix the verify function from crypto-util so that it can be used by the Verify component for ethereum addresses.

The problem was that the verify function expected a 36bytes address instead of the truncated 20bytes address used by ethereum.

I added two tests


jacogr (approved):
lgtm. Thank you very much. (Will update you when it gets into apps, hopefullyl later today)

PR #781, opened 2020-11-30
toJson gives a wrong address for ethereum type

Tbaut:
I found out about this one while doing https://github.com/polkadot-js/extension/issues/538
The address in the `pair` was the right one, but not the one in the Json.


packages/keyring/src/pair/toJson.spec.ts
```diff
@@ -37,4 +38,58 @@ describe('toJson', (): void => {
       }
     });
   });
+
+  const ethereumPair = new Keyring({ type: 'ethereum' });
+  const pairOpts = {
+    meta: {
+      genesisHash: '0xc8d32cd0020b9da3804ef726f876604b7fcf5a799e687221a959f2dba5c78f18',
```
Tbaut, line 45: Not sure that brings anything, this test is based on a real life example, so kept it as close to the reality.

packages/keyring/src/pair/toJson.spec.ts
```diff
@@ -37,4 +38,58 @@ describe('toJson', (): void => {
...
+  };
+
+  it('creates an unencoded output with no passphrase for ethereum account with ethereum type', (): void => {
+    const pair = ethereumPair.addFromUri(pairOpts.suri, pairOpts.meta, pairOpts.pairType as 'ethereum');
+
+    const json = pair.toJson();
+
+    expect(json).toMatchObject({
+      address: '0x5D0aa7d985DfF020174C907e3b57F5C8b5741013',
+      encoded: 'MFMCAQEwBQYDK2VwBCIEIHKHRqqkltcqvX4+97zbqofVujatYSmJ7HgsDvt4xWhmoSMDIQACqEZqFaw+I6dWa1sSwYCJn6wGo8VQ4/4GTV3EyCHIAiA=',
```
Tbaut, line 61: Note that the fix from this PR focuses on the address, I'm actually not confident regarding the other fields.

jacogr:
It is actually on purpose, in this case the `address` filed is not correctly named. The issue is that the hashing is one-way, so you cannot have access to the publicKey without decompressing the encoded part - which is actually very useful. By storing the full publicKey, you can however go to encoded address.

Tbaut:
Thanks for the explanation, for "ethereum" type, this `address` field actually contains the private key, which can be used to retrieve the address (with `keccak(publicKey).slice(-20)`).

PR #796, opened 2020-12-10
fix createPair for ethereum addresses

joelamouche:
in createPair, the TYPE_ADDRESS function for ethereum doesn't handle address properly


jacogr:
The crosscheck failure is fixed in master - yes, it is a bit of a "nothing-to-do-with-me" failure.

PR #808, opened 2020-12-15
detectPackage with import.meta.url

jacogr:


jacogr:
Cannot detect, replaced.

PR #834, opened 2021-01-13
Add Polymesh Ledger Integration

adamdossa:
Adds details needed for Polymesh ledger integration.

### DIFF
--- a/packages/networks/src/index.ts
+++ b/packages/networks/src/index.ts
@@ -133,6 +133,7 @@ const all: NetworkFromSubstrate[] = [
   {
     decimals: null,
     displayName: 'Polymath',
+    genesisHash: ['12fddc9e2128b3fe571e4e5427addcb87fcaf08493867a68dd6ae44b406b39c7'],
     network: 'polymath',
     prefix: 12,
     standardAccount: '*25519',
--- a/packages/networks/src/index.ts
+++ b/packages/networks/src/index.ts
@@ -131,13 +131,14 @@ const all: NetworkFromSubstrate[] = [
     website: 'http://laminar.network/'
   },
   {
-    decimals: null,
-    displayName: 'Polymath',
-    network: 'polymath',
+    decimals: [6],
+    displayName: 'Polymesh',
+    genesisHash: ['12fddc9e2128b3fe571e4e5427addcb87fcaf08493867a68dd6ae44b406b39c7'],
+    network: 'polymesh',
     prefix: 12,
     standardAccount: '*25519',
-    symbols: null,
-    website: null
+    symbols: ['POLYX'],
+    website: 'https://polymath.network/'
   },
   {
     decimals: null,
--- a/packages/networks/src/index.ts
+++ b/packages/networks/src/index.ts
@@ -133,7 +133,7 @@ const all: NetworkFromSubstrate[] = [
   {
     decimals: null,
     displayName: 'Polymath',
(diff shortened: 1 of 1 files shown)


adamdossa:
The three PRs (including this one) below are related:
https://github.com/polkadot-js/common/pull/834
https://github.com/polkadot-js/apps/pull/4397
https://github.com/polkadot-js/ui/pull/427

packages/networks/src/index.ts
```diff
@@ -131,13 +131,14 @@ const all: NetworkFromSubstrate[] = [
     website: 'http://laminar.network/'
   },
   {
-    decimals: null,
-    displayName: 'Polymath',
-    network: 'polymath',
+    decimals: [6],
+    displayName: 'Polymesh',
+    genesisHash: ['12fddc9e2128b3fe571e4e5427addcb87fcaf08493867a68dd6ae44b406b39c7'],
```
jacogr, line 136: As a first round, would just add the genesisHash which is needed for Ledger-operation.

The rest of the changes fail since this is checked against the ss58-registry on Substrate, so should map, see the crosscheck failure at https://github.com/polkadot-js/common/runs/1697045426?check_suite_focus=true#step:3:36 

(Background to this ... Atm this is duplicated - the idea is that that ss58-rregistry json be split into a seperate repo and be usable from both JS & Rust. At that point the only values here that will remain will be the genesisHashes, i.e. the fields not in the Substrate master edition)

jacogr (commented):
Thank you! One comment to just align with the upstream registry, the hash itself is 100%

packages/networks/src/index.ts
adamdossa: Thanks @jacogr - does this mean that changing the network name (`polymath` > `polymesh`) will always fail this crosscheck? If so, I wonder whether it would be possible to push the change through w/o this check?

If that's not possible I'll update to just include the `genesisHash` although ideally we'd also like to change the name at some point.

jacogr: Yes, just remove all the other changes until the ss58 registry on Substrate is updated, e.g. https://github.com/paritytech/substrate/blob/master/ss58-registry.json. (There is a nightly check that picks up discrepancies, so once merged there will apply the changes here)

All the JS stuff ignores it anyway, the important bits it does use is the network and then the genesisHash. (Things like the decimals and token is gets from the chain)

adamdossa: Thanks for the details. This PR is updated to just have the genesisHash. I'll create a PR against Substrate/.../ss58-registry.json with the other details.

packages/networks/src/index.ts
```diff
@@ -133,6 +133,7 @@ const all: NetworkFromSubstrate[] = [
   {
     decimals: null,
     displayName: 'Polymath',
+    genesisHash: ['12fddc9e2128b3fe571e4e5427addcb87fcaf08493867a68dd6ae44b406b39c7'],
```
jacogr, line 136:
```suggestion
    genesisHash: ['0x12fddc9e2128b3fe571e4e5427addcb87fcaf08493867a68dd6ae44b406b39c7'],
```

Needs the hex prefix.

(Sorry, should have seen this in the first rounf)

jacogr (approved):
Thank you.

jacogr:
Will bump all the downstream deps first thing in the morning along with your apps UI PR.

PR #847, opened 2021-01-21
Update ledger key to polymesh (from polymath)

adamdossa:
This aligns with the upstream ss58 registry and downstream ledger packages.

### DIFF
--- a/packages/hw-ledger/src/index.ts
+++ b/packages/hw-ledger/src/index.ts
@@ -24,7 +24,7 @@ const APPS: Record<string, (transport: Transport) => SubstrateApp> = {
   dock: newDockApp,
   kusama: newKusamaApp,
   polkadot: newPolkadotApp,
-  polymath: newPolymeshApp
+  polymesh: newPolymeshApp
 };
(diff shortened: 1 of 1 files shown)


jacogr (approved):
Thank you - actually was thinking about this when I updated the ss58 stuff but then it ended up being a stream of new updates and I seem to have dropped the ball.

adamdossa:
> Thank you - actually was thinking about this when I updated the ss58 stuff but then it ended up being a stream of new updates and I seem to have dropped the ball.

Thanks for the quick merge - do you know how long it takes for this to make it to the https://polkadot.js.org/apps page?

jacogr:
Going to apps is a manual update like for instance this PR https://github.com/polkadot-js/apps/pull/4462 - these type of PRs actually always bump to the latest of all polkadot-js stuff. 

So the latest packages from a repo are always published to npm as CI completes, but the use of that elsewhere requires manual intervention. 

There will be another bump-all PR a bit later today that will pull in latest from here and elsewhere.

Issue #773, opened 2020-11-17
Support Ethereum-type derivation paths


joelamouche:
In `common/packages/keyring/src/keyring.ts`
there is a comment : `FIXME Need to support Ethereum-type derivation paths`

Is this fixed yet? Are derivation path working for ethereum? If yes, what is the format?

Thanks in advance @jacogr

jacogr:
No. It only does the Substrate style. (Which is not compatible with Ethereum at all)

joelamouche:
Do you have an idea of the amount of work and complexity it would take to support that?

jacogr:
Implement bip-44, basically. The big issue I have with this is that I'm trying to get rid on not-frequently-used stuff, so generally not happy to just reach for JS libs and get it and all it's dependencies.

The only complexity is that Substrate always uses suri, so in the case of Ethereum the string (before stripping) won't be full compliant, i.e. 

seed `0x1234` with path `m/44'/0'/0'/1/1` would probably look like `0x1234/44'/0'/0'/1/1` in suri-like format before being used. 

(There is a use for bip-44 style derivation in the case of eg. Ledger as well, where ed keys are derived that way, even though it supports Polkadot/Kusama)

jacogr:
I'm quite happy to do this on the apps UI first, with hdkey libs, before adding it here.

So my concern with the crypto libs is stuff like this - https://github.com/cryptocoinjs/hdkey/blob/master/package.json#L39 - it actually took ages to get secp256k1 our and deduped (so elliptic is not used here only), both are absolutely massive.

So would suggest - 

- add via apps UI
- apps UI does derivation
- derived keys passed through the the keyring

It will also be less complex, just pull in libs and be done with it.

crystalin:
Hi @jacogr ,
What is the status on this?

jacogr:
It is not on the to-do-immediately list, so not a priority to get done. 

I have taken some looks at some of the derivations libs out there and it pulls in the wide whole world of dependencies in all cases I have seen. So atm have no clean, light solution for this - which is needed if it as to go into common.

crystalin:
I'd be happy to get your suggestion on what should be done here and what library you suggest and maybe @joelamouche will work to have it implemented. It is not an urgent matter but it feels weird being able to use the private key and not the mnemonic for our chain.

jacogr:
So HDKey is actually a good candidate - and the guys are solid.

[...]

PR #858, opened 2021-01-28
Implement first eth hd derivation test

joelamouche:
solves https://github.com/polkadot-js/common/issues/773

As per recommended by @jacogr : 
- [x] use hdkey as a base (we certainly on't need serialize, toJSON, etc.)
- [x] rip out the derives and do them in TS
- [x] adjust so it uses the libraries already available in util-crypto (all the stuff is there, eg. secp2561k1, hashing, ...)
- [x] make it work with the pairs (would need some adjustment)
- [x] add tests


joelamouche:
This is WIP but your input would be welcome @jacogr 
NB:
- the derivation in the test works but the derived address is wrong
- your implementation of secp256k1 doesn't have the required `publicKeyCreate` , `publicKeyConvert` and `publicKeyVerify`

joelamouche:
@jacogr Have you had time to take a look at this PR?
Thanks in advance

joelamouche:
@jacogr So I'm making progress and I now have a working version.

Question: I don't understand this test (packages/keyring/src/index.spec.ts):
```
it('allows for signing/verification', (): void => {
      const MESSAGE = stringToU8a('just some test message');
      const signer = keyring.createFromUri(PHRASE);
      const verifier = keyring.addFromJson(
        JSON.parse('{"address":"KWCv1L3QX9LDPwY4VzvLmarEmXjVJidUzZcinvVnmxAJJCBou","encoded":"U8qFEaghhmNV2PgFhjqzmhyUy37Ok7abfFU2MNsBd0sAgAAAAQAAAAgAAAA3+NniKogzNphiMNueB1X0sGA07B6CaXWfpXPx45iSXoTTprwzU5mOoSqUWO0GKHROI72LN+uJ8Yfv6Ll6JOOV3VPKfoVoFmYm+zDrrMPa0gk5E5kUuSijxADcE6zUrliPVr0Ix/qaghu5SJ7RtWDQLBf4Hp86SJ8Gg6gTSSk=","encoding":{"content":["pkcs8","ethereum"],"type":["scrypt","xsalsa20-poly1305"],"version":"3"},"meta":{}}')
      );

      const signature = signer.sign(MESSAGE);

      expect(verifier.verify(MESSAGE, signature)).toBe(true);
      expect(verifier.verify(new Uint8Array(), signature)).toBe(false);
    });
```

How can the verifier know wether the signature is good or not without the signer address?
Usually the verify function returns the signer address so I am confused about this test and the following

packages/util-crypto/src/mnemonic/toLegacySeed.ts
```diff
@@ -24,10 +24,17 @@ import { mnemonicValidate } from './validate';
...
  * ```
  */
-export function mnemonicToLegacySeed (mnemonic: string, password = '', onlyJs = false): Uint8Array {
-  assert(mnemonicValidate(mnemonic), 'Invalid bip39 mnemonic specified');
-
-  return isReady() && !onlyJs
-    ? bip39ToSeed(mnemonic, password)
-    : mnemonicToSeedSync(mnemonic, password).subarray(0, 32);
+export function mnemonicToLegacySeed(mnemonic: string, password = "", onlyJs = false, isEthereum: boolean): Uint8Array {
+  // IS THIS USED BY ANY OTHER BC THAN ETHEREUM? WE COULD SIMPLIFY THIS FUNCTION IF IT ISNT THE CASE
```
jacogr: Indeed. Actually Ledger as well, but for the time being that does not go via this function (just because is needs 64 bytes, not 32 - so the flag you added is actually what gets used in Ledger as well, but the way to correctly solve it is to have the byteLength passed, not the flag)

joelamouche: Ok thanks I will add the byte length.

Have you had the occasion of reviewing my comment above about the sign/verify tests? 

Thanks in advance

jacogr: Bt tricky, since atm if specified it needs to just ignore the wasm path (I actually did add it and then for some weird reason it didn't go into the commit healing with the Ledger HD - my mistake, I was switching between 2 branches there,  guess the final ended up being on the non-changed branch, i.e. it doesn't use this call there)

joelamouche:
@jacogr I understand you are busy but I just wanted to make sure you didn't miss my comment above about the sign/verify tests

jacogr:
1. Yes, the secp lib is only what is needed, i.e. no extra stuff based on previous usage. If extra is needed, it should be added. 

2. The key is here `https://github.com/polkadot-js/common/blob/master/packages/keyring/src/index.spec.ts#L332` - the signature contains the publicKey. so you only always need the signature & message to verify. (This is a true/false, it doesn't actually compare the publicKey, which it should actually do)

packages/util-crypto/src/secp256k1/privateKeyTweakAdd.ts
```diff
@@ -0,0 +1,63 @@
...
+  // PUBKEY_SERIALIZE: 'Public Key serialization error',
+  // PUBKEY_COMBINE: 'The sum of the public keys is not valid',
+  // SIG_PARSE: 'Signature could not be parsed',
+  // SIGN: 'The nonce generation function failed, or the private key was invalid',
+  // RECOVER: 'Public key could not be recover',
+  // ECDH: 'Scalar was invalid (zero or overflow)'
+};
+
+function isUint8Array(name: string, value: Uint8Array, length: number) {
+  assert(value instanceof Uint8Array, `Expected ${name} to be an Uint8Array`);
```
jacogr, line 27: `isU8a` from `@polkadot/util`

joelamouche: ok thx

packages/util-crypto/src/mnemonic/toLegacySeed.ts
joelamouche: ok

joelamouche:
NB: Derivation without private key (i.e. publickey derivation) will not be implemented in the hdkey class I'm submitting because publicKeyTweak is not trivial to implement and I don't see the use case in our ecosystem

packages/keyring/src/keyring.ts
```diff
@@ -6,7 +6,7 @@ import type { KeyringInstance, KeyringOptions, KeyringPair, KeyringPair$Json, Ke
 
 import { assert, hexToU8a, isHex, isUndefined, stringToU8a } from '@polkadot/util';
 import { base64Decode, decodeAddress, encodeAddress, ethereumEncode, keyExtractSuri, keyFromPath, mnemonicToLegacySeed, mnemonicToMiniSecret, naclKeypairFromSeed as naclFromSeed, schnorrkelKeypairFromSeed as schnorrkelFromSeed, secp256k1KeypairFromSeed as secp256k1FromSeed } from '@polkadot/util-crypto';
-import HDKey from '@polkadot/util-crypto/key/hdkey';
+import HDKey from '@polkadot/util-crypto/key/hdkeyEthereum';
```
jacogr, line 9: I would suggest moving this to the hd folder (that is where the Ledger derivation ended up)

joelamouche: Done! Ready for review

joelamouche:
@jacogr It would be great if you had the time to review this because this is pretty high priority for Moonbeam. Thanks in advance

jacogr:
Will take a look as soon as possible.

packages/keyring/src/keyring.ts
```diff
@@ -192,8 +193,20 @@ export class Keyring implements KeyringInstance {
...
+
+    if (type === 'ethereum') {
+      const key = HDKeyEth.fromMasterSeed(seed);
+      const child = key.derive(derivePath.substring(1));
+
+      if (child.publicKey && child.privateKey) {
+        derived = { publicKey: child.publicKey, secretKey: child.privateKey };
+      } else {
+        throw new Error('child derivation errored');
+      }
```
jacogr, line 206:
```suggestion
      assert(child.publicKey && child.privateKey, 'Unable to derive HD key from path');
      
      derived = { publicKey: child.publicKey, secretKey: child.privateKey };
```

Nothin major, just a suggestion.

packages/keyring/src/testing.ts
```diff
@@ -71,8 +94,9 @@ export const PAIRS: PairDef[] = [
  */
 export function createTestKeyring (options: KeyringOptions = {}, isDerived = true): KeyringInstance {
   const keyring = new Keyring(options);
+  const pairs:PairDef[] = (options.type && options.type === 'ethereum') ? PAIRSETHEREUM : PAIRSSR25519;
```
jacogr, line 97:
```suggestion
  const pairs: PairDef[] = (options.type && options.type === 'ethereum') ? PAIRSETHEREUM : PAIRSSR25519;
```

This issue with this one is that it won't actually ever get used, for instance in the apps UI. The test keyring has no type that is applied.

packages/util-crypto/src/hd/ethereum/index.ts
```diff
@@ -0,0 +1,224 @@
+// Copyright 2017-2021 @polkadot/util-crypto authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import * as crypto from 'crypto';
+// const Buffer = require('safe-buffer').Buffer;
```
jacogr, line 5:
```suggestion
```

packages/util-crypto/src/hd/ethereum/index.ts
```diff
@@ -0,0 +1,224 @@
...
+import { Buffer } from 'safe-buffer';
+
+// const assert = require('assert');
```
jacogr, line 8:
```suggestion
```

packages/util-crypto/src/hd/ethereum/index.ts
```diff
@@ -0,0 +1,224 @@
...
+// Bitcoin hardcoded by default, can use package `coininfo` for others
+const BITCOIN_VERSIONS = { private: 0x0488ADE4, public: 0x0488B21E };
+
+interface Versions{
+  private:number;
+  public:number;
+}
+
+// function to retype Buffer to ArrayBuffer in a controlled manner
+function bufferAsArrayBuffer (buffer:Buffer):ArrayBuffer {
```
jacogr, line 24:
```suggestion
function bufferAsArrayBuffer (buffer:Buffer): ArrayBuffer {
```

These are all nits. Really not sure why eslint doesn't force a space here. (really should between `:` and `'|` - which is used extensively here)

packages/util-crypto/src/hd/ethereum/index.ts
```diff
@@ -0,0 +1,224 @@
...
+// function to retype Uint8Array to Buffer in a controlled manner
+function bytesAsBuffer (bytes:Uint8Array):Buffer {
+  return bytes as unknown as Buffer;
+}
+
+export class HDKeyEth {
+  versions:Versions;
+  public depth:number;
+  index:number;
+  private _privateKey:Uint8Array|null;
```
jacogr, line 37: For private, use the private identifier, eg. `#privateKey: Uint8Array | null;`

packages/util-crypto/src/mnemonic/toLegacySeed.ts
```diff
@@ -24,10 +24,16 @@ import { mnemonicValidate } from './validate';
  * }
  * ```
  */
-export function mnemonicToLegacySeed (mnemonic: string, password = '', onlyJs = false): Uint8Array {
+export function mnemonicToLegacySeed (mnemonic: string, password = '', onlyJs = false, byteLength?: number): Uint8Array {
```
jacogr, line 27:
```suggestion
export function mnemonicToLegacySeed (mnemonic: string, password = '', onlyJs = false, byteLength: 32 | 64 = 32): Uint8Array {
```

packages/util-crypto/src/mnemonic/toLegacySeed.ts
```diff
@@ -24,10 +24,16 @@ import { mnemonicValidate } from './validate';
...
   assert(mnemonicValidate(mnemonic), 'Invalid bip39 mnemonic specified');
 
-  return isReady() && !onlyJs
-    ? bip39ToSeed(mnemonic, password)
-    : mnemonicToSeedSync(mnemonic, password).subarray(0, 32);
+  if (byteLength && byteLength === 32) {
```
jacogr, line 30: Would suggest an assert on the 32 | 64, e.g. `assert([32, 64].includes)byteLength), `Invalid seed length ${byteLength}, expected 32 or 64`);

packages/util-crypto/src/mnemonic/toLegacySeed.ts
```diff
@@ -24,10 +24,16 @@ import { mnemonicValidate } from './validate';
...
+    return isReady() && !onlyJs
+      ? bip39ToSeed(mnemonic, password)
+      : mnemonicToSeedSync(mnemonic, password).subarray(0, 32);
+  } else if (!byteLength || (byteLength && byteLength === 64)) {
```
jacogr, line 34:
```suggestion
  } else if (byteLength === 64) {
```

packages/util-crypto/src/mnemonic/toLegacySeed.ts
```diff
@@ -24,10 +24,16 @@ import { mnemonicValidate } from './validate';
...
+    return mnemonicToSeedSync(mnemonic, password);
+  } else {
+    throw new Error('wrong byte length for mnemonicToLegacySeed');
+  }
```
jacogr, line 38:
```suggestion
  }
```
With the suggested assert.

packages/util-crypto/src/secp256k1/tweakAdd.ts
```diff
@@ -0,0 +1,63 @@
...
+
+function isUint8Array (name: string, value: Uint8Array, length: number|number[]) {
+  assert(isU8a(value), `Expected ${name} to be an Uint8Array`);
+
+  if (length !== undefined) {
+    if (Array.isArray(length)) {
+      const numbers = length.join(', ');
+      const msg = `Expected ${name} to be an Uint8Array with length [${numbers}]`;
+
+      assert(length.includes(value.length), msg);
```
jacogr, line 25:
```suggestion
      assert(length.includes(value.length), `Expected ${name} to be an Uint8Array with length [${numbers}]`);
```

packages/util-crypto/src/secp256k1/tweakAdd.ts
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      const msg = `Expected ${name} to be an Uint8Array with length ${length}`;
+
+      assert(value.length === length, msg);
```
jacogr, line 29:
```suggestion
      assert(value.length === length, `Expected ${name} to be an Uint8Array with length ${length}`);
```

packages/util-crypto/src/secp256k1/tweakAdd.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const ec = new EC('secp256k1');
+const ecparams = ec.curve as {n:BN};
+
+const errors = {
+  PUBKEY_PARSE: 'Public Key could not be parsed',
+  TWEAK_ADD: 'The tweak was out of range or the resulted private key is invalid'
+};
+
+function isUint8Array (name: string, value: Uint8Array, length: number|number[]) {
```
jacogr, line 17: We can just use `isU8a` from `@polkadot/util` instead of this function. Then the use becomes `assert(isU8a(...), message)`

packages/util-crypto/src/key/extractSuri.ts
```diff
@@ -7,6 +7,7 @@ import { DeriveJunction } from './DeriveJunction';
 import { keyExtractPath } from './extractPath';
 
 export interface ExtractResult {
+  derivePath:string,
```
jacogr, line 10:
```suggestion
  derivePath: string,
```

jacogr (commented):
Seems sane against the test-cases - which is the most important.

I just would prefer actual derivation tests against vectors for the mnemonic + bip39 = known. (In the keyring this would make sense?). That just pulls it all together and proves the outputs are as excpected.

packages/util-crypto/src/hd/ethereum/index.ts
jacogr: Master now will auto-fix these when running with `yarn lint`.

packages/util-crypto/src/key/extractSuri.ts
jacogr: As above, master + `yarn lint` will fix it, so less nitpicks.

packages/keyring/src/testing.ts
joelamouche: >this one

What are you referring to? The type or PAIRSETHEREUM?

packages/util-crypto/src/hd/ethereum/index.ts
joelamouche: So #privateKey instead of _privateKey ?

packages/util-crypto/src/secp256k1/tweakAdd.ts
joelamouche: yes but isU8a doesn't check bytelength, does it?

jacogr: No. But it is still a simple `assert(isU8a(x) && x.length === y, message)`

joelamouche:
>I just would prefer actual derivation tests against vectors for the mnemonic + bip39 = known. 

I don't understand what you mean here

packages/util-crypto/src/secp256k1/tweakAdd.ts
joelamouche: ok sounds good, adding it

jacogr:
Sorry about being cryptic - to ensure it all ties together, a test somewhere that shows `derive(key, path)`

So this is where it may be tricky, since the `addFromUri` is generally `mnemonic//hard/soft` - so here we have an issue to tie it all in since we expect an `m` somewhere in there. (The suri is a full specified, it is not generally key + path arguments)

So basically - with known seed and path, using the keyring, do we get the expected outputs?

joelamouche:
I see. Well addFromUri is already tested in `packages/keyring/src/index.spec.ts` for the keyring.
If we want to test `key.derive` where key is HDKeyEth or basicaly those lines:

```
const key = HDKeyEth.fromMasterSeed(seed);
      const child = key.derive(derivePath.substring(1));
```
then we should test it on HDKeyEth's repo in util-crypto don't you think?

jacogr:
I don't see a specific m-path test like this - https://github.com/polkadot-js/common/pull/858/files#diff-7004d4e5bce0b271e81834800d01b1016c8206ae39113f3b4ed0e3d2dcc1f9fdR196 (granted, that one just uses the dev seed, so I have no idea how the fromUri would work with `abandon ... abandon` and `m/345'/0'/0'` - putting thos together yields `abandon ... abandonm/345'/0'/0'`)

joelamouche:
`abandon ... abandonm/345'/0'/0'` needs to be separated by a `/`
I added a test for HDKeyEth

Is that what you meant?

packages/keyring/src/keyring.ts
```diff
@@ -198,7 +198,7 @@ export class Keyring implements KeyringInstance {
     if (type === 'ethereum') {
       const key = HDKeyEth.fromMasterSeed(seed);
       const child = key.derive(derivePath.substring(1));
-
+      console.log('hehe',child.publicKey,child.privateKey )
```
jacogr, line 201: Comment that slipped-in.

jacogr:
The added tests look fine. Still would like to see a sample in the keyring itself via addUri as well - if for no other reason, just to have an example of use.

packages/keyring/src/keyring.ts
joelamouche: thanks

joelamouche:
I added a test for addUri

jacogr (approved):

PR #881, opened 2021-02-15
Add Eth dev address

joelamouche:
We also need a dev address for Ethereum type parachains.

Here is Gerald, the default prefunded address for moonbeam, that can be reused for any parachain that uses ethereum addresses


jacogr:
Closing. Gerald was added in https://github.com/polkadot-js/common/pull/858

PR #883, opened 2021-02-18
Pass publicKey through on verify

jacogr:
As per https://github.com/polkadot-js/common/pull/858#issuecomment-781289330


joelamouche:
nice

PR #894, opened 2021-02-25
Fix eth derivation

joelamouche:
Mainly problems related to 'toJson' and overall json conversion


jacogr:
ecdsa is broken with this. There is a test for the addition that i added - https://github.com/polkadot-js/common/pull/893/files#diff-7004d4e5bce0b271e81834800d01b1016c8206ae39113f3b4ed0e3d2dcc1f9fdR312-R325

joelamouche:
Nice! should be good now

packages/keyring/src/pair/index.ts
```diff
@@ -173,9 +173,8 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
       );
     },
     toJson: (passphrase?: string): KeyringPair$Json => {
-      // NOTE: in the case of Ethereum and Ecdsa, we encode the ful publicKey - to allow extraction. This will NOT match
-      // e.g. an Ethereum address which is non-reversible, so not used here)
-      const address = ['ecdsa', 'ethereum'].includes(type)
+      // For ecdsa and ethereum, address and publicKey are not the same (and not bijective). We want the publicKey here
+      const address = ['ethereum'].includes(type)
```
jacogr, line 177: We have dropped one. (It has the same issue, hence the original check and the check in master)

joelamouche: fixed

jacogr (approved):

PR #909, opened 2021-03-01
Jlm test from seed

joelamouche:
Adds a test for fromSeed to make sure it can be used in the apps.
See note


packages/util-crypto/src/secp256k1/keypair/fromSeed.spec.ts
```diff
@@ -14,7 +14,7 @@ const tests: Test[] = [
     'life fee table ahead modify maximum dumb such tobacco boss dry nurse',
     '0xf2360e871c830d397fe221382b503f07ddd8763df81a94bb2504390a2fb91f59',
     '0x036b0aa6beab469dd2b748a0ff5ddbe3d13df1e15c9d28a2aa057212994e127bea',
-    '0xae8e8fcacbaeb607bcdf0bbd7e615f2b4ef484ee54f19d68a7393fb6db2dd9cd'
+    '0xae8e8fcacbaeb607bcdf0bbd7e615f2b4ef484ee54f19d68a7393fb6db2dd9cd' // why are those here when the test only used the first 3 of the array?
```
jacogr, line 17:
```suggestion
    '0xae8e8fcacbaeb607bcdf0bbd7e615f2b4ef484ee54f19d68a7393fb6db2dd9cd'
```

jacogr: Not sure why the comment is here?

This is the converted seed, not checked since the priv/pub is enough.

joelamouche: No I was just wondering if we should delete those.

jacogr:
As per the issue logged in apps - you need to do the check against the keyring, since that is where it fails - the addFromUri path needs to handle it correctly.

jacogr:
As per the issue logged in apps - you need to do the check against the keyring, since that is where it fails - the addFromUri path needs to handle it correctly.

joelamouche:
@jacogr updated

packages/keyring/src/keyring.ts
```diff
@@ -194,7 +195,9 @@ export class Keyring implements KeyringInstance {
     }
 
     const derived = type === 'ethereum'
-      ? hdEthereum(seed, derivePath.substring(1))
+      ? isHex(phrase, 256)
```
jacogr, line 198: We already have this check above? If needed, rather set a flag.

packages/keyring/src/keyring.ts
```diff
@@ -194,7 +195,9 @@ export class Keyring implements KeyringInstance {
...
+        ? keypairFromSeed[type](seed) // for eth, if the private key is provided as suri, it must be derived only once
```
jacogr, line 199: In this case it means you cannot do derivations on the supplied key.

packages/keyring/src/keyring.ts
```diff
@@ -174,6 +174,7 @@ export class Keyring implements KeyringInstance {
       ? `${DEV_PHRASE}${_suri}`
       : _suri;
     const { derivePath, password, path, phrase } = keyExtractSuri(suri);
+
```
jacogr, line 177:
```suggestion
```

Unneeded space appeared here.

packages/keyring/src/keyring.ts
joelamouche: It means that derivation will only be done for a mnemonic.
I don't understand what you are saying about the UI

jacogr: On the UI, you have this -

image

So in the case of the raw seeds, since the keyring has no support for derivation, it should not have the derivation option.

joelamouche: ha yes indeed

joelamouche: let me take care of the common bump when its ready and I will make the necessary changes

joelamouche:
updated

jacogr (approved):

joelamouche:
nice, let me know when I can do the ui-keyring bump with UI modif in the apps

jacogr:
UI keyring does not need a bump. The apps UI uses resolutions for the common and api repos. (Generally UI only gets bumped when there are changes that needs code changes internally - and well on release day)

Issue #899, opened 2021-02-26
Add eth addresses to test dev addresses


joelamouche:
We want to add eth addresses when running a dev node (and prefund them in our genesis)
If I understand what you said here https://github.com/polkadot-js/apps/issues/4467#issuecomment-768233896 we can do that as long as we provide the secret key and the public key so that it can be created (and not derived) by createPair.
In fact, it looks like Gerald has already been added to the dev keyring for ethereum (as I have tested).
Can we add others addresses that would be considered canonic dev addresses for eth henceforth?

jacogr:
Does Gerald actually show up on the apps UI?

joelamouche:
We are flexible on what the standard should be.
We could go for "bottom drive obey lake curtain smoke basket hold race lonely fit walk" if that's what's standard.

<img width="1022" alt="Screenshot 2021-02-26 at 11 36 03" src="https://user-images.githubusercontent.com/16758383/109309895-29d70a00-7844-11eb-9627-0df9105eac5b.png">

jacogr:
If we stick to that, quite happy to adjust with whatever paths makes sense.

joelamouche:
>Does Gerald actually show up on the apps UI?

Not sure if it was saved or it showed up. What would prevent it from displaying? Is there a filter on pair type?

jacogr:
As far as I know, it just uses the defaults, doesn't try to pass through a type. So there may be some mangling required. Let's put it this way, if Alice/Bon/Charlie/... showed up, it is probably incorrect since your logic excluded those.

joelamouche:
I'm sorry I don't quite understand. Do you mind pointing to me where the dev pairs are declared and where they are loaded please?

jacogr:
Here - https://github.com/polkadot-js/common/blob/master/packages/keyring/src/testing.ts

The reason why the public/secret is already derived here (shown as hex, not suri) and not derived via mnemonic is that the mnemonic -> seed takes quite a while, i.e. it would have a delay of 1-2s on the UI layer as opposed to the approach here.

joelamouche:
Thanks. Where in the UI can I see this function being called. It's quite hard to understand without having access to ui-keyring's loadAll(). It doesn't seem to call createTestPairs in the compiled js version.
I need to verify that it's passing on the type

jacogr:
https://github.com/polkadot-js/ui/blob/master/packages/ui-keyring/src/Base.ts#L105-L121

[...]

PR #912, opened 2021-03-02
Add eth dev addresses

joelamouche:
Added the first 5 addresses generated by `ganache-cli --mnemonic "bottom drive obey lake curtain smoke basket hold race lonely fit walk"`
Gave them name that reminds they're eth addresses

resolves https://github.com/polkadot-js/common/issues/899


packages/keyring/src/testing.ts
```diff
@@ -73,6 +73,49 @@ export const PAIRSSR25519: PairDef[] = [
 ];
 
 export const PAIRSETHEREUM: PairDef[] = [
+  {
+    name: 'Alith',
+    publicKey: hexToU8a('0x02509540919faacf9ab52146c9aa40db68172d83777250b28e4679176e49ccdd9f'),
+    secretKey: hexToU8a('0x5fb92d6e98884f76de468fa3f6278f8807c48bebc13595d45af5bdc4da702133'),
+    seed: 'Alith',
```
jacogr, line 80: No sure the seed it correct? I'm assuming it is something like `m/44'/40'/0'/0`?

jacogr: (so in the case of the sr25519 is is basically the `//Alice` with the `//` removed (not quire correct removing it, but it made some other parts simpler, so it "sort-of stuck")

joelamouche: yeah no the seed is not correct. I added it for typing's sake. I guess I should add the seed derived from the mnemonic? Didn't think it would matter.

joelamouche: `m/44'/40'/0'/0` is not the seed, it's the derivation path used to derive the privatekey from the seed

jacogr: The "seed" here is used intechagably because of historic reasons -

- in the original version the "seed" was used as-is to derive keys
- in later versions it changed to `<known mnemonic>//<original seed>` (name was not changed)

So it does refer to the path.

joelamouche: good to know thanks. updated

jacogr (approved):
Thank you.

BTW, really like the names.

packages/keyring/src/testing.ts
jacogr: It should be changed and the actual sr* should really reflect the full path. Bug me each time I look at this file, but not enough to fix yet since it is "cosmetic". (Additionally the old seed-like derivation can be removed from the body)

PR #932, opened 2021-03-28
logger - wildcard include/exclude with DEBUG env

Lezek123:
Projects that make use of https://www.npmjs.com/package/debug may rely on `DEBUG` env to specify a set debuggers to use, which may include wildcards and exclusion groups (https://www.npmjs.com/package/debug#wildcards), for example:
```
DEBUG=*,-api:*
```
_(use all debuggers **except** those that begin with `api:`)_
```
DEBUG=api:*
```
_(use all debuggers that begin with `api:`)_

Currently it's not possible to exclude loggers created by `@polkadot` packages or include only one of the loggers, whithout including all others that start with the same set of characters.

This PR is a suggestion of how this problem may be addressed.

### DIFF
--- a/packages/util/src/logger.spec.ts
+++ b/packages/util/src/logger.spec.ts
@@ -118,6 +118,8 @@ describe('logger', (): void => {
   });
 
   it('does debug log when DEBUG partial specified', (): void => {
+    process.env.DEBUG = 'test*';
+
     l = logger('testing');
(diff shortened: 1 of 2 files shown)


jacogr (approved):
Thank you, this is great.

My sincere apologies for the long delay before taking a proper look. It has been on the list and staring me in the face.

PR #835, opened 2021-01-14
add hasLedgerSupport prop to networks

Tbaut:

PR #935, opened 2021-04-08
Add network for crust maxwell

yuhui1208:


jacogr:
The list here should only have what the up-stream registry has, e.g. https://github.com/paritytech/substrate/blob/master/ss58-registry.json

The addition is great, but won't be merged until it is accepted in the master Substrate list. 

(There is also a cron job running that checks this list against the substrate list - as soon as that splits into a separate repo, which is planned, a lot of the duplication here would be removed, only very specific information would be added here, e.g. genesis, but the bulk of the data would be de-duped)

yuhui1208:
> The addition is great, but won't be merged until it is accepted in the master Substrate list. The list here should only have what the up-stream registry has, e.g. https://github.com/paritytech/substrate/blob/master/ss58-registry.json
> 
> (There is also a cron job running that checks this list against the substrate list - as soon as that splits into a separate repo, which is planned, a lot of the duplication here would be removed, only very specific information would be added here, e.g. genesis, but the bulk of the data would be de-duped)
> 
> 42 is already allocated to the Substrate default. (Which actually means the checks for dupes needs to be adjusted)

Thanks for the reply. We want to support ledger in our apps. If we want to have these two fileds `genesisHash` and `hasLedgerSupport` in `packages/networks/src/index.ts`, how can we achieve this? We checked the code in ss58-registry.json before but don't kwow how to do it. Any light would be appreciated. 😄

jacogr:
The networks here needs to match the known networks in the registry - as it stands there is already duplication for all the basics (and it indeed has some extra fields, which will remain).

At this point really don't have a way of adding anything that is not in the master Substrate registry. (If they get added the cron checks will start failing)

yuhui1208:
Got it. According to #835 and #710, we can add extra fields into existing network. But we cannot create a new item in this json file. Am I right? Since 42 is alreay used by Substrate as the default ss58 format, so we would close this PR first.

jacogr:
100% - extra fields are fine and will be kept (even if we remove the duplication here and just work from the Substrate published ss58 registry - that will be split, as mentioned, just not done as of yet... so for the time being the exact same info is duplicated).

PR #947, opened 2021-04-14
Update defaults.ts with HydraDX re-genesis hash

jak-pan:


packages/networks/src/defaults.ts
```diff
@@ -14,6 +14,9 @@ export const knownGenesis: KnownGenesis = {
   edgeware: [
     '0x742a2ca70c2fda6cee4f8df98d64c4c670a052d9568058982dad9d5a7a135c5b'
   ],
+  hydradx: [
```
jacogr, line 17: Do you have the precious hash as well - accounts tied to the previous hash can also match to the new if both supplied.

jacogr: If this is the current, all ok.

packages/networks/src/defaults.ts
```diff
@@ -14,6 +14,9 @@ export const knownGenesis: KnownGenesis = {
...
+    '0x3d75507dd46301767e601265791da1d9cb47b6ebc94e87347b635e5bf58bd047'
```
jacogr, line 18:
```suggestion
    '0x3d75507dd46301767e601265791da1d9cb47b6ebc94e87347b635e5bf58bd047',
    '0x0ed32bfcab4a83517fac88f2aa7cbc2f88d3ab93be9a12b6188a036bf8a943c2'
```

Adding the current genesisHash from RPC.

jak-pan: Yes that's the previous one. Thanks!

jacogr (approved):

jacogr:
I will roll this out to the apps UI in a couple of hours - it may actually close the loop and just allow everything "to work", old or news.

- extension sets the hash, can be old or new
- apps UI will match hashes (either old or new)

So when the user uses the old or updates, he/she should already be able to set and use it in the extension after it.

(It should have been in already, but running a bit late, so can promise a couple of hours, will link to this issue)

jak-pan:
> I will roll this out to the apps UI in a couple of hours - it may actually close the loop and just allow everything "to work", old or news.
> 
> * extension sets the hash, can be old or new
> * apps UI will match hashes (either old or new)
> 
> So when the user uses the old or updated new, he/she should already be able to set and use it in the extension after it.
> 
> (It should have been in already, but running a bit late, so can promise a couple of hours, will link to this PR)

Thank you very much! You're quick 👀

PR #969, opened 2021-05-06
Jlm update for ethereum pair addition

joelamouche:
When injecting accounts, we only have the address and not the publicKey.
Therefore, we need to support adding a key from address for ethereum addresses


packages/keyring/src/pair/index.ts
```diff
@@ -113,23 +113,28 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
   };
 
   const encodeAddress = (): string => {
-    const raw = TYPE_ADDRESS[type](publicKey);
+    // Need to be able to add from an eth address
+    if (type === 'ethereum' && publicKey.length === 20) {
+      return u8aToHex(publicKey);
+    } else {
+      const raw = TYPE_ADDRESS[type](publicKey);
```
jacogr, line 120: This is the logic that should be in the `TYPE_ADDRESS[type](...)` function - if that receives 20 bytes as input into that function, then treat it as such.

joelamouche: TYPE_ADDRESS[type](...) dosn't leave room for flexibility. Please make a PR with your suggested changes

jacogr: Possibly after parachains have launched on Kusama, don't have much playing time atm.

joelamouche: This is not playing it's important work. This change is very important to Moonbeam. We have been trying to get MetaMask compatibility for a long time now. Please include this change that is tested and we can refactor it together when you have more time

jacogr: Ok. So... I'm not trying to get you upset here.

You asked for a PR from me, I mentioned where my time is spent. I sadly cannot drop what I'm busy with atm. 

I understand it is important to your team, not diminishing the effort at all, it has been great and moved things along greatly. However, anything that stands in the way to parachains on Polkadot so we can pull all of this effort together, is "playtime" which don't have as much as I wish.

So... for best-effort, would suggest we cleanup this PR here, remove the duplication and use the functions as intended - even if they may need to change to cater for this. If you want me to attack this rather than changing this PR, as you asked for here, it will need to wait until there is a gap.

jacogr: I mean it is a change of `ethereum: (p: Uint8Array) => keccakAsU8a(secp256k1Expand(p)),` to `ethereum: (p: Uint8Array) => p.length === 20 ? p : keccakAsU8a(secp256k1Expand(p)),`

1. function used as intended
2. no duplication where used

joelamouche: Ok sorry I thought it would be easier this way. I pushed a commit to use the function as intended

packages/keyring/src/pair/index.ts
```diff
@@ -44,7 +44,13 @@ const TYPE_SIGNATURE = {
 const TYPE_ADDRESS = {
   ecdsa: (p: Uint8Array) => p.length > 32 ? blake2AsU8a(p) : p,
   ed25519: (p: Uint8Array) => p,
-  ethereum: (p: Uint8Array) => keccakAsU8a(secp256k1Expand(p)),
+  ethereum: (p: Uint8Array) => {
+    if (p.length === 20) {
+      return (p);
+    } else {
+      return hexToU8a(ethereumEncode(keccakAsU8a(secp256k1Expand(p))));
```
jacogr, line 51: Since Ethereum addresses are with the checksum encoding, I would not change this. It adds value in the ecosystem.

joelamouche: Not change what?

jacogr: The encoding. So we updated all address outputs here to be non-checksummed. In the Ethereum ecosystem the checksums do add value. If we want non checksum addresses (which we certainly don't), the correct place would be to encode it without. In this case however, we don't want to do this.

1. the address output should output the U8a - only a simple `p.length === 20 ? p : keccakAsU8a(secp256k1Expand(p))`
2. the encodeaddress should use ethereumEncode

So effectively we will probably only have the line change in 1 as code-changes (for length 20, return as-is, else keep the status quo) and then your new tests to ensure it is working

packages/keyring/src/pair/index.ts
```diff
@@ -44,7 +44,13 @@ const TYPE_SIGNATURE = {
...
   ecdsa: (p: Uint8Array) => p.length > 32 ? blake2AsU8a(p) : p,
   ed25519: (p: Uint8Array) => p,
-  ethereum: (p: Uint8Array) => keccakAsU8a(secp256k1Expand(p)),
+  ethereum: (p: Uint8Array) => {
+    if (p.length === 20) {
+      return p;
+    } else {
+      return hexToU8a(ethereumEncode(keccakAsU8a(secp256k1Expand(p))));
+    }
+  },
```
jacogr, line 53:
```suggestion
  ethereum: (p: Uint8Array) => p.length === 20 ? p : keccakAsU8a(secp256k1Expand(p)),
```

This seems like the only code change required (before the addition of the specific test as per the first commit)

jacogr: See https://github.com/polkadot-js/common/compare/jg-eth-address?expand=1 for the one line change + tests (lifted from here, only addition is the removal of the toLowerCase on the address)

packages/keyring/src/pair/index.ts
joelamouche: Sounds good, I pushed a commit in that sens

packages/keyring/src/pair/index.ts
```diff
@@ -115,9 +115,7 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
   const encodeAddress = (): string => {
     const raw = TYPE_ADDRESS[type](publicKey);
 
-    return type === 'ethereum'
-      ? ethereumEncode(raw)
-      : toSS58(raw);
+    return type === 'ethereum' ? (ethereumEncode(raw)) : toSS58(raw);
```
jacogr, line 118:
```suggestion
    return type === 'ethereum'
      ? ethereumEncode(raw)
      : toSS58(raw);
```

The original here is fine, not 100% sure why the ()'s slipped in.

packages/keyring/src/pair/index.ts
```diff
@@ -127,8 +125,9 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     get addressRaw (): Uint8Array {
       const raw = TYPE_ADDRESS[type](publicKey);
 
-      return type === 'ethereum'
-        ? raw.slice(-20)
+      return type === 'ethereum' &&
+          raw.length > 20
+        ? (raw).slice(-20)
```
jacogr, line 130:
```suggestion
      return type === 'ethereum'
        ? raw.slice(-20)
```

Original fine as well, if 20 bytes still works exactly as expected.

jacogr (commented):
Looks good, just 2 small comments around reverting to the previous versions as indicated. Happy to merge after that, thank you.

jacogr (approved):
Thank you!

PR #984, opened 2021-05-21
added equilibrium slip44; added equilibrium genesis hash

alexdniep:


alexdniep:
@jacogr am I correct that for knownLedger value I should use value from here https://github.com/satoshilabs/slips/blob/master/slip-0044.md?

jacogr:
Indeed. So the code here doesn't apply the needed hard derivation, so it expects it to be already accounted for.

jacogr:
Tests are failing since it also needs an entry here - https://github.com/polkadot-js/common/blob/master/packages/hw-ledger/src/defaults.ts#L12 (to map the app to the actual app to construct from the Ledger package)

alexdniep:
> Tests are failing since it also needs an entry here - https://github.com/polkadot-js/common/blob/master/packages/hw-ledger/src/defaults.ts#L12 (to map the app to the actual app to construct from the ZondaX Ledger package)

Got it, only what should we use for CLA value like
`
const CLA = {  KUSAMA: 0x99,  POLKADOT: 0x90,  POLYMESH: 0x91,  DOCK: 0x92,  CENTRIFUGE: 0x93,  EDGEWARE: 0x94,}`
`

jacogr:
The CLA is used internally by the ZondaX package. So importing the correct app, as exposed there, just takes care of it. (So then when it gets to the Ledger is uses the correct CLA and also signing will obviously use the correct on-device genesisHash as part of the signing. Basically these are hardcoded to those apps on the hardware side)

jacogr (approved):
Thank you.

PR #1006, opened 2021-06-07
Add Edgeware ledger app

raykyri:
Adds Ledger support for Edgeware, which should be supported by the Zondax library as of https://github.com/Zondax/ledger-substrate-js/pull/35.


jacogr (approved):
Thank you.

PR #1008, opened 2021-06-10
Compute HDKD within function(s)

kwingram25:
Small adjustment to compute HDKD const(s) within relevant deriveHard functions instead of at file level. This is to correct an issue found running Jest jsdom tests for canvas-ui.

https://github.com/paritytech/canvas-ui-v2/pull/22#issuecomment-858814917


jacogr (commented):
This form is explicitly done to not have to do the allocation on each call. 

It is done is a number of places, not just here - when there is a constant conversion, do it outside the function. There is no need to do the allocation on each run.

PR #1032, opened 2021-06-23
Fix tojson pair for 20 address

joelamouche:
In the case where the publickey is a 20bytes string, it means that the address is passed directly and should be copied into address


packages/keyring/src/pair/index.ts
```diff
@@ -176,8 +176,9 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     toJson: (passphrase?: string): KeyringPair$Json => {
       // NOTE: For ecdsa and ethereum, the publicKey cannot be extracted from the address. For these
       // pass the hex-encoded publicKey through to the address portion of the JSON (before decoding)
+      // unless the publicKey is already an address
       const address = ['ecdsa', 'ethereum'].includes(type)
-        ? u8aToHex(secp256k1Compress(publicKey))
+        ? publicKey.length === 21 ? u8aToHex(secp256k1Compress(publicKey)) : u8aToHex(publicKey)
```
jacogr, line 181:
```suggestion
        ? publicKey.length === 20
          ? u8aToHex(publicKey)
          : u8aToHex(secp256k1Compress(publicKey))
```

jacogr, line 181: I am not sure in which cases `=== 21` would be true here? The publicKey would be longer than that and the address is exact 20.

joelamouche: Are you sure it's not 21 with the prefix? This worked when copying on the apps.
Just like 33 is for 32 bytes

jacogr: So the raw Uint8Array is 20 bytes for Ethereum addresses. 

The ecdsa public is indeed 33 (normal Substrate is 32)

joelamouche: Ok makes sens

joelamouche:
@jacogr updated

jacogr (approved):
Thank you.

PR #1047, opened 2021-07-07
Extending Substrate network registry with `Heiko` and `Parallel`

### LINKED ISSUE #1046: SS58 registry is outdated
cc @polkadot-js/notifications

Discrepancies detected between https://github.com/paritytech/substrate/blob/master/ss58-registry.json and https://github.com/polkadot-js/common/blob/master/packages/networks/src/substrate.ts

Check the nightly cron output (or via `yarn networks:crosscheck` locally) and run `yarn networks:sync` to update. The output as found from the test includes:

Number of entries mismatched:: Expected 65 found 63

Missing entries found: [
  "Heiko (110)",
  "Parallel (172)"
]

adamzerella:
Resolves #1046

Copied from the registry networks listed in https://github.com/polkadot-js/common/blob/master/packages/networks/src/substrate.ts

@tomusdrw not sure who to add as the reviewer here

### DIFF
--- a/packages/networks/src/substrate.ts
+++ b/packages/networks/src/substrate.ts
@@ -577,5 +577,23 @@ export const knownSubstrate: KnownSubstrate[] = [
     standardAccount: '*25519',
     symbols: ['BSX'],
     website: 'https://bsx.fi'
+  },
+  {
+    decimals: [12],
+    displayName: 'Heiko',
+    network: 'heiko',
+    prefix: 110,
+    standardAccount: '*25519',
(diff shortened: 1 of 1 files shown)


jacogr (commented):
The ordering here is different as to the original. As per the comments here, the correct way to sync would be to run `yarn networks:sync` and the script will take care of it.

(Sadly the CI checks don't run on PRs, otherwise it should have picked up the manual editing - if it did run, it would fail on all PRs while the registries are out of sync)

PR #1057, opened 2021-07-10
Add CFG Ledger Support

mikiquantum:


mikiquantum:
@jacogr Did I miss any other place so the ledger dropdown option shows up in the Apps settings?

jacogr:
The tests actually check that everything is covered - so if they pass, you will be fine.

jacogr (approved):
Thank you.

Issue #1068, opened 2021-07-26
[feature request] add in encrypt/decrypt method for common/keyring


RoyTimes:
Hi, per discussion in https://github.com/polkadot-js/common/issues/1067 The project we are working on needs to either obtain private key from browser extension **OR** enable the browser extension to encrypt/decrypt message with a recipient's public key and the private key from the keyring. 

I guess it's better not to mess with exposing the private key. Therefore, I guess the second option would be optimal. 

A few questions before we start drafting a PR:
1. Do you think encrypt/decrypt are good functionalities to be included? I believe Metamask offers some ways of doing so. 
2. "It might be tricky to add in support for all elliptic curves: ed25519 is covered by tweetnacl and we work with the lib a lot; secp256k1 is implemented by "elliptic" and offers an ECDH and should be good; sr25519 is tricky and I wonder if I should mess with the Rust wasm code. It's good enough for us to only have ed25519, or use ed25519 for all encrypt/decrypt. I wonder how much compromise is acceptable. " (EDITED: I might be foggy on this. Both sr25519 & ed25519 should use the same curve for ECDH)
3. For browser extension, we will also propose a PR related to this but won't open a new issue. 
4. For browser extension, I wonder if it should pop up a new window similar to how "signing" works, or the extension could do it quietly. I don't think there's much risk for user to encrypt data (unlike signing), but might be required to pop up a window when decrypting. 

Let me know your thoughts. Tests will be included.

jacogr:
Closed in #1070

PR #1070, opened 2021-07-26
Add in ability to encrypt/decrypt msg with keyring

RoyTimes:
Ref: https://github.com/polkadot-js/common/issues/1068

- 1 dependency added to 'util-crypto' - `ed2curve` by the author of tweetnacl.js 
- tests included
- covers sr25519 & ed25519, secp256k1 to be added.

### DIFF
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -147,6 +147,8 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     decodePkcs8,
     decryptMessage: (encryptedMessageWithNonce: string | Uint8Array, senderPublicKey: string | Uint8Array): Uint8Array | null => {
       assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
+      assert(type !== 'ecdsa', 'Secp256k1 not supported yet');
+      assert(type !== 'ethereum', 'Secp256k1 not supported yet');
 
       return naclOpen(
         u8aToU8a(encryptedMessageWithNonce.slice(24, encryptedMessageWithNonce.length)),
@@ -169,6 +171,9 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     },
     encryptMessage: (message: string | Uint8Array, recipientPublicKey: string | Uint8Array, _nonce?: Uint8Array): Uint8Array => {
       assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
+
+      assert(type !== 'ecdsa', 'Secp256k1 not supported yet');
+      assert(type !== 'ethereum', 'Secp256k1 not supported yet');
 
       const { nonce, sealed } = naclSeal(u8aToU8a(message), convertSecretKeyToCurve25519(secretKey), convertPublicKeyToCurve25519(u8aToU8a(recipientPublicKey)), _nonce);
 
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -171,9 +171,7 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     },
     encryptMessage: (message: string | Uint8Array, recipientPublicKey: string | Uint8Array, _nonce?: Uint8Array): Uint8Array => {
       assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
-
-      assert(type !== 'ecdsa', 'Secp256k1 not supported yet');
-      assert(type !== 'ethereum', 'Secp256k1 not supported yet');
+      assert(!['ecdsa', 'ethereum'].includes(type), 'Secp256k1 not supported yet');
 
       const { nonce, sealed } = naclSeal(u8aToU8a(message), convertSecretKeyToCurve25519(secretKey), convertPublicKeyToCurve25519(u8aToU8a(recipientPublicKey)), _nonce);
 
--- a/packages/keyring/src/pair/index.ts
+++ b/packages/keyring/src/pair/index.ts
@@ -147,8 +147,7 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     decodePkcs8,
     decryptMessage: (encryptedMessageWithNonce: string | Uint8Array, senderPublicKey: string | Uint8Array): Uint8Array | null => {
       assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
-      assert(type !== 'ecdsa', 'Secp256k1 not supported yet');
-      assert(type !== 'ethereum', 'Secp256k1 not supported yet');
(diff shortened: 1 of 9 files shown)


packages/util-crypto/src/nacl/convertKey.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2017-2021 @polkadot/util-crypto authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+/* eslint-disable */
```
jacogr, line 4: Would not do this at all. Rather add a small `ed2curve.d.ts`

packages/keyring/src/pair/index.ts
```diff
@@ -145,6 +145,21 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
...
+      const _senderPublicKey = convertPublicKeyToCurve25519(u8aToU8a(senderPublicKey));
+
+      const message = naclOpen(
+        u8aToU8a(encryptedMessageWithNonce.slice(24, encryptedMessageWithNonce.length)),
+        u8aToU8a(encryptedMessageWithNonce.slice(0, 24)),
+        _senderPublicKey,
+        _secretKey
+      );
+
+      return message;
```
jacogr, line 161:
```suggestion
      assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');

      return naclOpen(
        u8aToU8a(encryptedMessageWithNonce.slice(24, encryptedMessageWithNonce.length)),
        u8aToU8a(encryptedMessageWithNonce.slice(0, 24)),
        convertPublicKeyToCurve25519(u8aToU8a(senderPublicKey)),
        convertSecretKeyToCurve25519(secretKey)
      );

      return message;
```

packages/keyring/src/pair/index.ts
```diff
@@ -157,6 +172,16 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
...
     },
+    encryptMessage: (message: string | Uint8Array, recipientPublicKey: string | Uint8Array, _nonce?: Uint8Array): Uint8Array => {
+      assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
+
+      const _secretKey = convertSecretKeyToCurve25519(secretKey);
+      const _recipientPublicKey = convertPublicKeyToCurve25519(u8aToU8a(recipientPublicKey));
+
+      const { nonce, sealed } = naclSeal(u8aToU8a(message), _secretKey, _recipientPublicKey, _nonce);
+
+      return u8aConcat(nonce, sealed);
```
jacogr, line 183:
```suggestion
      assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');

      const { nonce, sealed } = naclSeal(u8aToU8a(message), convertSecretKeyToCurve25519(secretKey), convertPublicKeyToCurve25519(u8aToU8a(recipientPublicKey)), _nonce);

      return u8aConcat(nonce, sealed);
```

packages/util-crypto/src/nacl/index.ts
```diff
@@ -15,3 +15,7 @@ export { naclVerify } from './verify';
 export { naclBoxKeypairFromSecret } from './box/fromSecret';
 export { naclOpen } from './open';
 export { naclSeal } from './seal';
+export {
+  convertSecretKeyToCurve25519,
+  convertPublicKeyToCurve25519
+} from './convertKey';
```
jacogr, line 21:
```suggestion
export { convertSecretKeyToCurve25519, convertPublicKeyToCurve25519 } from './convertKey';
```

jacogr (commented):
Funky hack to make this operational in sr/ed/eth.

Some small comments.

The only think I feel is missing is possibly explicit tests for various keypair types -

- Ethereum
- sr25519
- ed25519

packages/util-crypto/src/nacl/convertKey.ts
RoyTimes: nvm. there is a `@types/ed2curve`! Removing eslint-disable.

RoyTimes:
I have included test cases for cross sr & ed. I'm fairly positive if the tests can be added and work as expected in a similar fashion as the test for cross sr & ed. However, I'm not sure if it's the right cryptographic thing to do when forcibly converting private keys of the same size from x25519 to secp256k1. Two options here: add in an assert to reject usage on secp256k1 curve; OR just do the conversion and compose test cases for them. What do you think?

jacogr:
As a first-step, rather play safe and assert them away. 

We can re-visit if need be or add an alternative approach for them (which would probably be the correct approach here).

RoyTimes:
Done. 

On a better non-hacky approach: I have no doubt that all of ed25519, sr25519 and secp256k1 can do ECDH; on a worse side, the library for secp256k1 supports ECDH but not encryption/decryption and I might have to implement a portion of the cryptographic myself. On a better side, introduce new dependencies that is well trusted to do ed25519 sr25519 ECDH. Tweetnacl does not have it built in. For sr25519, that will involve messing with the Rust implementation. 

Long story short, it's a pretty heavy load of work to be done. I'm happy to revisit this when I'm more available.

When this is merged, I'll jump to `extension`.

packages/keyring/src/pair/index.ts
```diff
@@ -145,6 +145,18 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     },
     // eslint-disable-next-line sort-keys
     decodePkcs8,
+    decryptMessage: (encryptedMessageWithNonce: string | Uint8Array, senderPublicKey: string | Uint8Array): Uint8Array | null => {
+      assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
+      assert(type !== 'ecdsa', 'Secp256k1 not supported yet');
+      assert(type !== 'ethereum', 'Secp256k1 not supported yet');
```
jacogr, line 151:
```suggestion
      assert(!['ecdsa', 'ethereum'].includes(type), 'Secp256k1 not supported yet');
```

packages/keyring/src/pair/index.ts
```diff
@@ -157,6 +169,16 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     encodePkcs8: (passphrase?: string): Uint8Array => {
       return recode(passphrase);
     },
+    encryptMessage: (message: string | Uint8Array, recipientPublicKey: string | Uint8Array, _nonce?: Uint8Array): Uint8Array => {
+      assert(!isLocked(secretKey), 'Cannot encrypt with a locked key pair');
+
+      assert(type !== 'ecdsa', 'Secp256k1 not supported yet');
+      assert(type !== 'ethereum', 'Secp256k1 not supported yet');
```
jacogr, line 176:
```suggestion
      assert(!['ecdsa', 'ethereum'].includes(type), 'Secp256k1 not supported yet');
```

jacogr (approved):
LGTM. That you very much. 

Like mentioned, this is an interesting approach, certainly not something I thought about to get over the sr25519-doesn't-have-that-in-the-libs (yet).

PR #1077, opened 2021-08-11
Update default.ts with SORA network

NoodleSploder:
Update default.ts with SORA network


jacogr:
Should be combined with #1078

NoodleSploder:
Failed check.  I'll update the code and combine with other commit

PR #1089, opened 2021-08-23
add nodle ledger specs

ETeissonniere:
mutate the variables that will allow the polkadot js app frontend to enable ledger support for the nodle chain


jacogr (approved):
Thank you.

PR #1114, opened 2021-09-22
Improve ethereum tests

joelamouche:
While reviewing how keyring derives address from suri (mnemonic+derivation path), I found minor things to correct in the ethereum tests

Note: 60 is the canonical derivation code for ethereum


packages/keyring/src/suri.spec.ts
```diff
@@ -5,10 +5,12 @@
 
 import { u8aToHex } from '@polkadot/util';
 import { cryptoWaitReady } from '@polkadot/util-crypto';
+import { KeypairType } from '@polkadot/util-crypto/types';
```
jacogr, line 8:
```suggestion
import type { KeypairType } from '@polkadot/util-crypto/types';
```

Would just move it first where the types generally go.

packages/keyring/src/testing.ts
```diff
@@ -77,42 +77,42 @@ export const PAIRSETHEREUM: PairDef[] = [
...
-    seed: "m/44'/40'/0'/0",
+    seed: "m/44'/60'/0'/0",
     type: 'ethereum'
   },
   {
     name: 'Baltathar',
     publicKey: hexToU8a('0x033bc19e36ff1673910575b6727a974a9abd80c9a875d41ab3e2648dbfb9e4b518'),
     secretKey: hexToU8a('0x8075991ce870b93a8870eca0c0f91913d12f47948ca0fd25b49c6fa7cdbeee8b'), // nosemgrep
-    seed: "m/44'/40'/0'/0",
+    seed: "m/44'/60'/0'/0",
```
jacogr, line 87: This won't break where it is already used?

jacogr (commented):
2 tiny comments.

packages/keyring/src/testing.ts
joelamouche: actually the seed here is just a filler because the private key has already been derived so let me remove it

joelamouche:
@jacogr updated

packages/keyring/src/testing.ts
jacogr: Ahh, yes, you are right.  It was basically there as a hint of how to get to it.

jacogr (approved):

PR #1136, opened 2021-10-21
Update dock networks to latest specs

SamHellawell:
Dock had a change from a PoA network to PoS not too long ago, the PoA network is no longer online. Latest spec names and genesis hashes are provided in this PR

### DIFF
--- a/packages/hw-ledger/src/defaults.ts
+++ b/packages/hw-ledger/src/defaults.ts
@@ -10,7 +10,7 @@ import { newBifrostApp, newCentrifugeApp, newDockApp, newEdgewareApp, newEquilib
 export const ledgerApps: Record<string, (transport: Transport) => SubstrateApp> = {
   bifrost: newBifrostApp,
   centrifuge: newCentrifugeApp,
-  'dock-mainnet': newDockApp,
+  'dock-pos-mainnet': newDockApp,
   edgeware: newEdgewareApp,
   equilibrium: newEquilibriumApp,
   genshiro: newGenshiroApp,
--- a/packages/networks/src/defaults.ts
+++ b/packages/networks/src/defaults.ts
@@ -24,7 +24,7 @@ export const knownIcon: KnownIcon = {
 export const knownLedger: KnownLedger = {
   bifrost: 0x00000314,
   centrifuge: 0x000002eb,
-  'dock-mainnet': 0x00000252,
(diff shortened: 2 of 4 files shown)


SamHellawell:
@jacogr do you know what the effects would be on removing the `dock-mainnet` spec names like this PR already does? i think for most polkadot apps it will be fine, but im not sure about the extension. if a user of the extension currently has an account under `dock-mainnet` and then this update is applied (and they download it) - will the user still be able to see that account in the extension in order to export it so they can reimport for the PoS network?

jacogr:
In the extension, it matches on genesisHash. So what will happen is that both will show up, old and new. It won't be injected/used in the page if the genesisHash does not match.

SamHellawell:
Sounds good, we don't need it to be injected on the page but we do need for them to be able to export/reimport their accounts or change the network to the latest chains

jacogr:
Will merge as soon as https://github.com/paritytech/ss58-registry/pull/31 is in

jacogr:
(Small conflict, may be related to the previous sync with the ss58-registry done today)

jacogr:
Please rebase with master - the network details are not duplicated anymore (the 8.4.1 release will send this into the wild, https://github.com/polkadot-js/common/pull/1408) so the upstream ss58registry is used directly.

However the genesis hashes here still needs an update. (We can actually merge that part while we are still waiting on the upstream registry to be updated, i.e. just the genesis part here is fully mergeable)

SamHellawell:
Have rebased with latest master. Can you clarify what you mean by just the genesis part is fuilly mergeable?

Made a change to [ss58registry.test.json](https://github.com/polkadot-js/common/pull/1136/files#diff-02023e214a5c878f02afd2c7cf15538974bda3caa5c67a92d5c46478b8640499), can revert this if thats what you mean.

The spec name also updated, its no longer `dock-mainnet` on the nodes. So switching just genesis hash and not spec name, if thats what you meant, might not work.

jacogr:
Ahh, this https://github.com/polkadot-js/common/pull/1136/files#diff-02023e214a5c878f02afd2c7cf15538974bda3caa5c67a92d5c46478b8640499 is checked against the upstream `@substrate/ss58-registry` package, so yes, it will fail.

So we still need to wait for the PR in that repo to be merged.

jacogr:
Closing for now. Feel free to -re-open when adjusted.

PR #1163, opened 2021-11-06
Use `Buffer` for hex encoding/decoding and concatenation on Node.js

nazar-pc:
Turns out iterating byte after byte in Node.js is not the fastest thing in the world, why not using `Buffer` that can offload this to native code (potentially with SIMD, etc.).


jacogr:
This is not actually for Node - this is for all. For all non-Node we are really trying to make sure we drop Buffer. For Node-only, rather take this approach with an exports entry.

PR #1188, opened 2021-11-15
Use @noble/hashes

jacogr:
This requires BigInt ... so we are slightly held hostage by older versions of RN.

@Tbaut Do you know the state of `BigInt` support on RN? i.e. do these work in iOs/Android?

```js
// test 1 (syntax)
const a = 1n + 2n;

// test 2 (does it have BigInt?)
const b = BigInt(1) + BigInt(2);
```

Trying to figure out what to do with this, i.e. 

- specific (old-code) overrides for react-native (like we do elsewhere, RN has an own export & tested path)
- only allow react-native to use the WASM path (only caveat is crypto await before using)
- revisit every now and again until support appears and then go ahead here
- other?

I have been searching and it seem that iOs is probably less problematic than Android (which may be on a very old version of JSC). Then again, the info I could find has been very scarce.


packages/util-crypto/src/mnemonic/bip39.ts
```diff
@@ -40,9 +40,9 @@ function bytesToBinary (bytes: number[]): string {
 function deriveChecksumBits (entropyBuffer: Uint8Array): string {
   return bytesToBinary(
     Array.from(
-      hash.sha256().update(entropyBuffer).digest()
+      sha256(entropyBuffer)
     )
-  ).slice(0, (entropyBuffer.length * 8) / 32);
+  ).slice(0, (entropyBuffer.length) * 8 / 32);
```
jacogr, line 45:
```suggestion
  ).slice(0, (entropyBuffer.length * 8) / 32);
```

Tbaut:
Sorry about the late reply, I'm afk. Will try it out within the next week and let you know.

jacogr:
No issues. When you have time - at least for the hashing think I do have a fallback plan in-place, for stuff like secp256k1, not really.

Tbaut:
From what I could search there are shims to go around the fact that rn doesn't support Bigint as suggested here https://github.com/facebook/react-native/issues/28492

jacogr:
The shins won’t do the trick at all. The issue is that it cannot polyfill stuff like 1n + 2n. In this Case it is also a 3rd party that introduces it, so cannot use the Google polyfill. 

But anyway, let‘s see. I thin iOS has support, but android with JSC is always lagging.

packages/util-crypto/src/detectPackage.ts
```diff
@@ -2,9 +2,13 @@
...
+import { detectPackage, hasBigInt, packageInfo as utilInfo } from '@polkadot/util';
 import { packageInfo as randomInfo } from '@polkadot/x-randomvalues';
 
 import { packageInfo } from './packageInfo';
 
 detectPackage(packageInfo, typeof __dirname !== 'undefined' && __dirname, [netInfo, utilInfo, randomInfo]);
+
+if (!hasBigInt) {
+  console.warn('Your environment does not support JS BigInt, legacy fallbacks would be used for certain operations');
+}
```
jacogr, line 14:
```suggestion
```

packages/util-crypto/src/detectPackage.ts
```diff
@@ -2,9 +2,13 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { packageInfo as netInfo } from '@polkadot/networks/packageInfo';
-import { detectPackage, packageInfo as utilInfo } from '@polkadot/util';
+import { detectPackage, hasBigInt, packageInfo as utilInfo } from '@polkadot/util';
```
jacogr, line 5:
```suggestion
import { detectPackage, packageInfo as utilInfo } from '@polkadot/util';
```

jacogr:
@Tbaut https://blog.logrocket.com/whats-new-in-react-native-0-65/ - it lists `BigInt.prototype.toLocaleString` under the Android updates.

jacogr:
Following the approach with fallbacks when `typeof BigInt === 'undefined'`

PR #1189, opened 2021-11-15
Use @noble/secp256k1

jacogr:
After #1188 

Still has some gaps and as per the above PR, it needs BigInt, which may have some support issues. In this case it is very problematic, unlike hashing, there is no WASM fallbacks.

EDIT: The above comment refers to no WASM, interfaces are being added in https://github.com/polkadot-js/common/pull/1276 (this needs to pull those in)


packages/util/src/assert.ts
```diff
@@ -4,6 +4,7 @@
 import { isFunction } from './is/function';
 import { isNull } from './is/null';
 import { isUndefined } from './is/undefined';
+import { hasBigInt } from './has';
```
jacogr, line 7:
```suggestion
```

packages/util/src/assert.ts
```diff
@@ -42,3 +43,7 @@ export function assertReturn <T> (value: T | undefined | null, message: string |
 
   return value;
 }
+
+export function assertBigInt (): void {
+  assert(hasBigInt, 'This operation is not supported in your environment, BigInt support is required');
+}
```
jacogr, line 49:
```suggestion
```

PR #1291, opened 2021-12-07
Add eth uri keyring test

joelamouche:
I am trying to figure out https://github.com/polkadot-js/extension/issues/907
The signature seems to be working correctly and so is the payload 🤔


jacogr:
Additional tests are always good, it ensure that we don't break stuff along the way.

jacogr (approved):

Issue #1314, opened 2021-12-12
Encryption/Decryption not working with sr25519 keypair


LaurentTrk:
Hi all !
This issue is the continuation of the closed #1124 about the support of sr25519 keypair for the [Encryption/Decryption feature](https://github.com/polkadot-js/common/pull/1070).     

After several tests, I cannot get it work, as the `ed2curve.convertPublicKey` used to convert the public key does not work for sr25519 public key.  

The [Ristretto decoding](https://ristretto.group/formulas/decoding.html) of the public key used in [Schnorrkel](https://github.com/w3f/schnorrkel) is not the same as in ed25519.

I managed to make encryption/decryption between 2 sr22519 keypairs using [Ristretto255.js](https://github.com/novifinancial/ristretto255-js), but I can't do it between one ed25519 and one sr25519.

**Could anyone help me ?**

Thanks !

LaurentTrk:
Thanks @jacogr for your reply !  
I don't know what to think about it...

This issue is closely related to [this feature on the extension](https://github.com/polkadot-js/extension/issues/928)(which will be obsolete if this one does not work), and I think it was the plan of @RoyTimes when he first worked on this encryption feature.

RoyTimes:
@LaurentTrk @jacogr Let me look into it. I thought such case is covered with the original tests? I can change the code so that it appends the 32 bytes Curve25519 public key of the sender to the beginning of the message.

LaurentTrk:
Thanks @RoyTimes !  
Yes, `sr25519` tests seems covered, but it appears that coming from an `ed25519` and convert it to `sr255159` (the way it's done in the tests) works, whereas using a native `sr25519` does not.

The test code I am using:
```javascript
const keyring = new Keyring();

const senderPair = keyring.createFromUri(mnemonicGenerate(), { name: 'first pair' }, 'sr25519');
const receiverPair = keyring.createFromUri(mnemonicGenerate(), { name: 'second pair' }, 'sr25519');

const message = stringToU8a('This is a test.');
const encryptedMessage = senderPair.encryptMessage(message, receiverPair.publicKey);
const decryptedMessage = receiverPair.decryptMessage(encryptedMessage, senderPair.publicKey);

const isMatch = u8aToString(message) === u8aToString(decryptedMessage);

// Verify that the decrypted message matches the original message
console.log(`Does the decrypted message match the original message? ${isMatch}`);
```

[...]

PR #1320, opened 2021-12-14
Include sender's public key in the encrypted message 

RoyTimes:
This fixes https://github.com/polkadot-js/common/issues/1314 and added the code posted by @LaurentTrk in the tests.

### DIFF
--- a/packages/keyring/src/pair/index.spec.ts
+++ b/packages/keyring/src/pair/index.spec.ts
@@ -19,10 +19,10 @@ describe('pair', (): void => {
   });
 
   const SIGNATURE = new Uint8Array([80, 191, 198, 147, 225, 207, 75, 88, 126, 39, 129, 109, 191, 38, 72, 181, 75, 254, 81, 143, 244, 79, 237, 38, 236, 141, 28, 252, 134, 26, 169, 234, 79, 33, 153, 158, 151, 34, 175, 188, 235, 20, 35, 135, 83, 120, 139, 211, 233, 130, 1, 208, 201, 215, 73, 80, 56, 98, 185, 196, 11, 8, 193, 14]);
-  const ENCRYPTED = new Uint8Array([185, 28, 49, 143, 5, 202, 59, 85, 28, 78, 148, 63, 61, 45, 253, 117, 23, 140, 124, 129, 210, 170, 11, 117, 161, 185, 152, 242, 44, 29, 98, 64, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 132, 59, 53, 85, 227, 40, 107, 48, 234, 198, 238, 157, 238, 224, 235, 179, 75, 153, 241, 142, 254]);
+  const ENCRYPTED = new Uint8Array([64, 66, 233, 248, 187, 191, 112, 203, 133, 158, 33, 5, 234, 154, 152, 150, 19, 65, 210, 172, 115, 143, 130, 42, 16, 117, 169, 43, 212, 21, 198, 85, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 72, 5, 150, 194, 248, 135, 140, 215, 78, 199, 246, 215, 166, 202, 187, 191, 188, 77, 134, 80, 196]);
 
-  // the last byte is changed from 254 -> 253
-  const ENCRYPTED_CHANGED = new Uint8Array([185, 28, 49, 143, 5, 202, 59, 85, 28, 78, 148, 63, 61, 45, 253, 117, 23, 140, 124, 129, 210, 170, 11, 117, 161, 185, 152, 242, 44, 29, 98, 64, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 132, 59, 53, 85, 227, 40, 107, 48, 234, 198, 238, 157, 238, 224, 235, 179, 75, 153, 241, 142, 253]);
+  // the last byte is changed from 196 -> 197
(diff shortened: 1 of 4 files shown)


packages/keyring/src/pair/index.spec.ts
```diff
@@ -171,20 +154,36 @@ describe('pair', (): void => {
...
     ).toEqual(message);
   });
 
+  it('passes the test on issue#1314', (): void => {
+    // Ref: https://github.com/polkadot-js/common/issues/1314
+
+    const keyring = new Keyring();
+
+    const senderPair = keyring.createFromUri(mnemonicGenerate(), { name: 'first pair' }, 'ed25519');
+    const receiverPair = keyring.createFromUri(mnemonicGenerate(), { name: 'second pair' }, 'ed25519');
```
LaurentTrk, line 177: Unfortunately, the issue #1314 talks about `sr25519`, the previous code was ok with `ed25519`.

RoyTimes: Uh. Sorry! Forget to change this in testing. Updating soon.

packages/keyring/src/pair/index.ts
```diff
@@ -175,7 +175,7 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
 
       const { nonce, sealed } = naclSeal(u8aToU8a(message), convertSecretKeyToCurve25519(secretKey), convertPublicKeyToCurve25519(u8aToU8a(recipientPublicKey)), nonceIn);
```
LaurentTrk, line 176: This code, especially the `convertPublicKeyToCurve25519()` call, is likely to fail (depending on the bytes generated)

```javascript
    const sr25519Pair = keyring.createFromUri(mnemonicGenerate(), { name: 'an sr25519 pair' }, 'sr25519');
    const sr25519Curve = convertPublicKeyToCurve25519(u8aToU8a(sr25519Pair.publicKey));
    console.log(sr25519Curve);  //    <------ returns a null value
```

RoyTimes: My mistake. I forgot to replace ed25519 to sr25519 in testing. I have done the testing and this fix is not correct yet.

RoyTimes:
I introduced a new member of Keypair called `encryptionPublicKey` that is derived from `secretKey.slice(0, 32)` and the public key is calculated on Curve25519. I'm mostly certain that this is a secure operation but leaves the extension or application built on top of this harder, as you cannot simply use the address to send encrypted message anymore. Converting sr25519 pub key to curve25519 is risky (to implement). Doing ECDH directly on sr25519 is possible but harder than converting and is a nightmare to deal with and require users to be on the same KeyringType. So far, I can't think of a quicker and low-risk solution. 

Also, it seems to me that decode/encode wallet from secrets might be problematic that might worth attention, because the initial tests passed on cross-usage of ed25519/sr25519. On the tests proposed by @LaurentTrk the key conversion from `pairA.publicKey` to curve25519 should failed, because `ed2curve` cannot correctly calculate keys on sr25519. But it didn't fail on the original test. Therefore, I just assumed that it works. But apparently, I was wrong. Let me know if I'm, thinking about this the wrong way. 

```javascript
  it('validates encryption between ed25519 & sr25519 key pairs', (): void => {
    const PASS = 'testing';
    const encodedA = keyring.alice.encodePkcs8(PASS);
    const encodedB = keyring.bob.encodePkcs8(PASS);

    const pairA = createPair({ toSS58, type: 'sr25519' }, { publicKey: keyring.alice.publicKey });
    const pairB = createPair({ toSS58, type: 'sr25519' }, { publicKey: keyring.bob.publicKey });

    pairA.decodePkcs8(PASS, encodedA);
    pairB.decodePkcs8(PASS, encodedB);
    const message = new Uint8Array([0x61, 0x62, 0x63, 0x64, 0x65]);

    // alice-sr25519 -> bob-ed25519
    const encrypted1 = pairA.encryptMessage(message, keyring.bob.publicKey);

    // alice-sr25519 -> bob-sr25519
    const encrypted2 = pairA.encryptMessage(message, pairB.publicKey);

    // alice-ed25519 -> bob-sr25519
    const encrypted3 = keyring.alice.encryptMessage(message, pairB.publicKey);

    // decrypt: Bob-ed25519 - use alice-ed25519 pubkey
    expect(
      keyring.bob.decryptMessage(encrypted3, keyring.alice.publicKey)
    ).toEqual(message);

    // decrypt: Bob-ed25519 - use alice-sr25519 pubkey
    expect(
      keyring.bob.decryptMessage(encrypted1, pairA.publicKey)
    ).toEqual(message);

    // decrypt: Bob-sr25519 - use alice-sr25519 pubkey
    expect(
      pairB.decryptMessage(encrypted2, pairA.publicKey)
    ).toEqual(message);
  });
```

Also, I think it's now possible to do encryption/decryption on secp256k1 now if doing it this way. Will test it out this week.

LaurentTrk:
Thanks @RoyTimes !  
As a non cryptographer, it sounds to me that we are trying to force a square peg into a round hole...  

I would prefer to go for a well documented approach like [ECIES](https://cryptobook.nakov.com/asymmetric-key-ciphers/ecies-public-key-encryption), which seems compatible both with [ed25519](https://github.com/phayes/ecies-ed25519) and sr25519 (using the `schnorrkelAgreement` already available in `wasm-crypto`).

In this approach, we only need one keypair (the one of the receiver), so we don't care about cross wallet exchange.

I can start to work on this, any thought/idea is welcome :)

jacogr:
Sorry about the delay in looking, a bit behind on reviews - I have a note of getting to this one.

jacogr (commented):
I do believe I prefer the newer approach in https://github.com/polkadot-js/common/pull/1331 since it does rely on a more standard approach to exchange information. 

It does also suffer from some "please communicate this information before using", in this case it would be the non-generally-available encrypted public key, in the other case it is the explicit keypair type, which is most probably a lesser evil.

I would prefer to converge on one solution between the two, intead of playing lord-and-master and having to decide myself between 2 different solutions for the same underlying issue.

jacogr:
Closing, removing this functionality in https://github.com/polkadot-js/common/pull/1762

Issue #929, opened 2021-03-19
Public key / Private key encryption using polkadot lib?


SachinSWN:
I'm building a substrate blockchain backed by polkadot js. While working on a task I have a requirement of public/private key encryption and decryption. for ex: Bob wants to send Alice an encrypted email. To do this, Bob takes Alice's public key and encrypts his message to her. Then, when Alice receives the message, she takes the private key that is known only to her in order to decrypt the message from Bob. Does anyone know how this is done through Polkadot js library? Help will be appreciated!

jacogr:
For ed25519 pairs there is a encrypt/decrypt pairing -

- https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/nacl/encrypt.ts
- https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/nacl/decrypt.ts

It is not available for sr25519 (see https://github.com/polkadot-js/common/issues/633), nor ecdsa

SachinSWN:
Thank You..it was really helpful!

SachinSWN:
without NaclBoxKeyPair, can we encrypt/decrypt message using the normal public key and private key pair which a nemonic already have ?

jacogr:
No, there is no sr25519 exposed interfaces, see https://github.com/polkadot-js/common/issues/633

Issue #633, opened 2020-07-09
Asymmetric encryption and decryption support


AndreasGassmann:
I can see that there is support for encryption and decryption for [nacl](https://github.com/polkadot-js/common/tree/master/packages/util-crypto/src/nacl) using secretbox. But it looks like it is symmetric because the `encrypt` method needs the secret key.

Are there plans to add asymmetric encryption (encrypt with public key, decrypt with public key) as well?

I'm specifically looking for support for a `sr25519` keypair. But I haven't found any other implementations that do this. Are there plans to do this, if even possible?

The use case is basically that this would allow you to encrypt a message for a specific address owner.

jacogr:
There are the `naclSeal` & `naclOpen` methods which have been contributed. I believe they fit the bill.

AndreasGassmann:
Thanks for the response. I have to admit that I don't have a deep understanding of cryptography, but what exactly is the difference between `sr25519` and `ed25519`? If I understood correctly, then `sr25519` has improved signatures in comparison to `ed25519`, but does the encryption/decryption still work the same way?

In our library I was able to successfully add encryption/decryption for `ed25519` using libsodium, but when I try to use the method with the `sr25519` keypair I get an error (I don't remember which one).

I didn't see the `naclSeal` and `naclOpen` methods, thanks for pointing them out. However I'm a bit confused because it looks like the `naclOpen` method needs the senders public key.

```
export default function naclOpen (sealed: Uint8Array, nonce: Uint8Array, senderBoxPublic: Uint8Array, receiverBoxSecret: Uint8Array): Uint8Array | null {
  return nacl.box.open(sealed, nonce, senderBoxPublic, receiverBoxSecret) || null;
}
```

Also in the test, it seems that there is a step missing where the sender sends over his public key:

```
    const sender = naclKeypairFromString('sender');
    const receiver = naclKeypairFromString('receiver');

    // Make id key for identifying each other
    const senderIdKey = fromPath(sender, keyExtractPath('//1//1//1//1').path, 'ed25519');
    const receiverIdKey = fromPath(receiver, keyExtractPath('//2//2//2//2').path, 'ed25519');

    // Receiver sends encrypting public key to receive message to decrypt with his private key
    const receiverIdBoxKey = naclBoxKeypairFromSecret(receiverIdKey.secretKey);

    console.log(`Receiver sends receiver's public key to sender ${receiverIdBoxKey.publicKey.toString()}`);

[...]

PR #1331, opened 2021-12-20
Asymmetric encryption/decryption feature

LaurentTrk:
Hi all,

This work is an alternative way to add the asymmetric encryption/decryption feature than the one [already implemented](https://github.com/polkadot-js/common/pull/1070), which [does not work with sr25519](https://github.com/polkadot-js/common/issues/1314) (see [workaround proposal](https://github.com/polkadot-js/common/pull/1320) by @RoyTimes).

It works more like the [eth_decrypt function](https://docs.metamask.io/guide/rpc-api.html#eth-decrypt) from Metamask, where only one public key/private key pair is involved : you can encrypt data with the public key of the recipient, and only this recipient can decrypt the data with his private/secret key (vs the already implemented solution using [nacl.box](https://github.com/dchest/tweetnacl-js#public-key-authenticated-encryption-box) which requires 2 keypairs).

### What has been done ?

- Add `encrypt/decrypt`  methods to `util-crypto\sr25519` following [ECIES](https://cryptobook.nakov.com/asymmetric-key-ciphers/ecies-public-key-encryption)
- Add `encrypt/decrypt` methods to `util-crypto\ed25519` using existing encryption feature, but with an ephemeral key (similar as ECIES)
- Add `encrypt` method to `keyring` (_encrypt data without the need of a keypair_)
- Add `encrypt/decrypt` methods to  `keyringPair`

### Implementation details
Following SR25519 Encryption/Decryption following Elliptic Curve Integrated Encryption Scheme (ECIES)  
See https://cryptobook.nakov.com/asymmetric-key-ciphers/ecies-public-key-encryption

The algorithms used were chosen among those already used in this library.

1. **Ephemeral Key generation**
      Generate new keypair using the wasm `sr25519KeypairFromSeed` function, with a random seed from
      `mnemonicGenerate`
2. **Key Agreement**
      Use wasm `sr25519Agree` function between the generated ephemeral private key and the recipient public key
3. **Key Derivation**
      Use `pbkdf2` (random salt is generated, default 2048 rounds) to derive a new secret from the previous step output
      The derived secret is split into :
      - MAC key (first 32 bytes)
      - encryption key (last 32 bytes)
4. **Encryption**
      Use `nacl.secretbox` api symmetric encryption (xsalsa20-poly1305) to encrypt the message
      with the encryption key generated at step 3.
      A nonce (24 bytes) is randomly generated.
5. **MAC Generation**
      HMAC SHA256 (using the MAC key from step 3) of the concatenation of the encryption nonce, ephemeral public key and encrypted message

The encrypted message is the concatenation of the following elements :
- nonce (24 bytes) : random generated nonce used for the symmetric encryption (step 4)
- keyDerivationSalt (32 bytes) : random generated salt used for the key derivation (step 3)
- public key (32 bytes): public key of the ephemeral generated keypair (step 1)
- macValue (32 bytes): mac value computed at step 5
- encrypted (remaining bytes): encrypted message (step 4)

---
Feel free to share your comments and feedbacks :pray: 

**Nota :** Some folks were interested in this feature:
- [Asymmetric encryption and decryption support](https://github.com/polkadot-js/common/issues/633)
- [Public key / Private key encryption using polkadot lib?](https://github.com/polkadot-js/common/issues/929)
- [Feature Request: Asymmetric Encryption](https://github.com/polkadot-js/extension/issues/691)

### DIFF
--- a/packages/keyring/src/keyring.ts
+++ b/packages/keyring/src/keyring.ts
@@ -5,7 +5,7 @@ import type { EncryptedJsonEncoding, Keypair, KeypairType } from '@polkadot/util
 import type { KeyringInstance, KeyringOptions, KeyringPair, KeyringPair$Json, KeyringPair$Meta } from './types';
 
 import { assert, hexToU8a, isHex, isUndefined, stringToU8a } from '@polkadot/util';
-import { base64Decode, decodeAddress, ed25519PairFromSeed as ed25519FromSeed, encodeAddress, ethereumEncode, hdEthereum, keyExtractSuri, keyFromPath, mnemonicToLegacySeed, mnemonicToMiniSecret, secp256k1PairFromSeed as secp256k1FromSeed, sr25519PairFromSeed as sr25519FromSeed } from '@polkadot/util-crypto';
+import { base64Decode, decodeAddress, ed25519PairFromSeed as ed25519FromSeed, encodeAddress, encrypt as cryptoEncrypt, ethereumEncode, hdEthereum, keyExtractSuri, keyFromPath, mnemonicToLegacySeed, mnemonicToMiniSecret, secp256k1PairFromSeed as secp256k1FromSeed, sr25519PairFromSeed as sr25519FromSeed } from '@polkadot/util-crypto';
 
 import { DEV_PHRASE } from './defaults';
 import { createPair } from './pair';
@@ -295,5 +295,15 @@ export class Keyring implements KeyringInstance {
    */
   public toJson (address: string | Uint8Array, passphrase?: string): KeyringPair$Json {
     return this.#pairs.get(address).toJson(passphrase);
+  }
+
+  /**
+   * @name encrypt
+   * @summary Encrypt a message using the given publickey
+   * @description Returns the encrypted message of the given message using the public key.
+   * The encrypted message can be decrypted by the corresponding keypair using keypair.decrypt() method
+   */
+  public encrypt (message: string | Uint8Array, recipientPublicKey: string | Uint8Array, recipientKeyType?: KeypairType): Uint8Array {
+    return cryptoEncrypt(message, recipientPublicKey, recipientKeyType || this.type);
   }
 }
--- a/packages/keyring/src/pair/index.spec.ts
+++ b/packages/keyring/src/pair/index.spec.ts
@@ -237,17 +237,20 @@ describe('pair', (): void => {
 
   it('allows encrypt/decrypt with ed25519 keypair', (): void => {
     const message = new Uint8Array([0x61, 0x62, 0x63, 0x64, 0x65]);
-    const encryptedMessage = keyring.alice.encrypt(message);
+    const encryptedMessage = keyring.alice.encrypt(message, keyring.bob.publicKey);
 
-    expect(keyring.alice.decrypt(encryptedMessage)).toEqual(message);
+    expect(keyring.bob.decrypt(encryptedMessage)).toEqual(message);
+    expect(keyring.alice.decrypt(encryptedMessage)).toEqual(null);
   });
 
   it('allows encrypt/decrypt with sr25519 keypair', (): void => {
-    const sr25519KeyPair = new Keyring().createFromUri(mnemonicGenerate(), { name: 'sr25519 pair' }, 'sr25519');
+    const aliceSR25519KeyPair = new Keyring().createFromUri(mnemonicGenerate(), { name: 'sr25519 pair' }, 'sr25519');
+    const bobSR25519KeyPair = new Keyring().createFromUri(mnemonicGenerate(), { name: 'sr25519 pair' }, 'sr25519');
     const message = new Uint8Array([0x61, 0x62, 0x63, 0x64, 0x65]);
-    const encryptedMessage = sr25519KeyPair.encrypt(message);
+    const encryptedMessage = aliceSR25519KeyPair.encrypt(message, bobSR25519KeyPair.publicKey);
 
-    expect(sr25519KeyPair.decrypt(encryptedMessage)).toEqual(message);
+    expect(bobSR25519KeyPair.decrypt(encryptedMessage)).toEqual(message);
+    expect((): Uint8Array | null => aliceSR25519KeyPair.decrypt(encryptedMessage)).toThrow("Mac values don't match");
   });
 
   it('allows encrypt for an ed25519 keypair', (): void => {
@@ -264,6 +267,17 @@ describe('pair', (): void => {
     const encryptedMessage = keyring.encrypt(message, sr25519KeyPair.publicKey, sr25519KeyPair.type);
 
     expect(sr25519KeyPair.decrypt(encryptedMessage)).toEqual(message);
+  });
+
+  it('fails to encrypt when locked', (): void => {
+    const aliceSR25519KeyPair = createPair({ toSS58, type: 'sr25519' }, { publicKey: keyring.alice.publicKey });
+    const bobSR25519KeyPair = createPair({ toSS58, type: 'sr25519' }, { publicKey: keyring.alice.publicKey });
+    const message = new Uint8Array([0x61, 0x62, 0x63, 0x64, 0x65]);
+
+    expect(aliceSR25519KeyPair.isLocked).toEqual(true);
+    expect((): Uint8Array =>
+      aliceSR25519KeyPair.encrypt(message, bobSR25519KeyPair.publicKey)
+    ).toThrow('Cannot encrypt with a locked key pair');
   });
 
   describe('ethereum', (): void => {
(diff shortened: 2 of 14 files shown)


h4x3rotab:
Generally good. However I have two suggestions:

1. We need better documentation. The interface exposed is `encrypt` and `decrypt`, which hides all the details behind. We should at least include the basic algorithm and message format used in the comments, including:

    - How the ephemeral key pair is generated?
    - How the key is derived? Does it follow any crypto standard?
    - Which symmetric encryption algorithm and parameter is used?

    This is important because once the ecosystem grows, polkadotjs will not be the only library to implement the cryptographic. It's important to have clear guide for other SDKs to implement the standard with common crypto libraries. The later requirement also implies we should our best to stick to popular crypto standard and algorithm.

2. Expose the key agreement interface. Per ECIES, under the hood the symmetric key is generated by two parties. In this PR, it only addresses a special case where one party is represented by an ephemeral key. It assumes all the developer has to use an ephemeral key. However, I believe the key agreement is very useful as well. An idea is to leave the key agreement exposed, and then add a new function to generate the ephemeral key. So the two use cases can be fully covered. An example API design:

     - `keypair.encrypt(otherPartyPubkey, plainMessage)`
     - `keypari.decrypt(otherPartyPubkey, encryptedMessage)`

    The `otherPartyPubkey` can be easily generated by:

    ```js
    const mnemonic = mnemonicGenerate();
    const pair = keyring.addFromUri(mnemonic, { name: 'first pair' }, 'sr25519');
    ```

LaurentTrk:
Thanks @h4x3rotab !
I will add implementation details shortly.

Regarding your proposal on exposing the agreement, I would like to propose the 2 options (ephemeral key as ECIES and 2nd party key), depending on the way you call the `encrypt` method:

- `keypair.encrypt(otherPartyPubkey, plainMessage)` : use the private key of the key pair (no need of an ephemeral keypair)
- `keyring.encrypt(plainMessage)` : generate an ephemeral keypair (as specified in ECIES)

In both options, the `decrypt()` method does not need the public key argument, as the public key used (either from an ephemeral key or a 2nd party key) is embedded in the encrypted message.

revick:
Thanks for this work @LaurentTrk - I'm not able to add any coding or encryption contributions but from a usability point `eth_decrypt` requires the MetaMask user to decrypt each message by approving the popup dialogue; it would simpler for a user to permit decryption once and allow the dApp to encrypt decrypt at will from thereon. Although that may be relevant to the UI implementation following your work here.

gdethier:
Hello @LaurentTrk, thanks a lot for this contribution! I reviewed the code and IMHO, everything is fine (documentation, style, tests). Is there something preventing you to change the status of this PR to pending? If not, I suggest that you do so in order for maintainers to review it as well.

LaurentTrk:
Thanks for your comment @gdethier ;)   
I was expecting more reviews, but it seems that we are ready now...

packages/keyring/src/pair/index.ts
```diff
@@ -144,6 +144,14 @@ export function createPair ({ toSS58, type }: Setup, { publicKey, secretKey }: P
     },
     // eslint-disable-next-line sort-keys
     decodePkcs8,
+    decrypt: (encryptedMessage: HexString | string | Uint8Array): Uint8Array | null => {
+      assert(!isLocked(secretKey), 'Cannot decrypt with a locked key pair');
+      assert(!['ecdsa', 'ethereum'].includes(type), 'Secp256k1 not supported yet');
+
+      return type === 'ed25519'
+        ? ed25519Decrypt(u8aToU8a(encryptedMessage), { publicKey, secretKey })
+        : sr25519Decrypt(u8aToU8a(encryptedMessage), { publicKey, secretKey });
```
gdethier, line 153: In above 2 lines, no need to pass-in the public key, only the secret key is needed to decrypt the message.

gdethier (approved):
I just left a small suggestion.

gdethier:
Hello @jacogr ! Did you already have a look at this PR?

We would really love to see this feature integrated as we rely on it in this fork of the Polkadot JS extension which brings encryption/decryption: https://github.com/LaurentTrk/extension/tree/DecryptionFeature (we are actually waiting for this PR to be merged before creating the one on the extension).

If there is anything we can do to make your life easier, please do not hesitate to tell.

3yekn:
I am also anxiously awaiting this feature, into this library and the corresponding one on the extension. thank you!

jacogr:
Will bump it up the overflowing task queue. Sorry for the long time getting here, it _is_ on the list, but not a lot of progress seems to be made on my side getting through.

packages/keyring/src/keyring.ts
```diff
@@ -296,4 +296,14 @@ export class Keyring implements KeyringInstance {
...
   }
+
+  /**
+   * @name encrypt
+   * @summary Encrypt a message using the given publickey
+   * @description Returns the encrypted message of the given message using the public key.
+   * The encrypted message can be decrypted by the corresponding keypair using keypair.decrypt() method
+   */
+  public encrypt (message: string | Uint8Array, recipientPublicKey: string | Uint8Array, recipientKeyType?: KeypairType): Uint8Array {
+    return cryptoEncrypt(message, recipientPublicKey, recipientKeyType || this.type);
```
jacogr, line 307: Generally users have no idea even what their own keypair types are, let alone somebody else knowing what a specific user has.

This means that the addition of the `recipientKeyType` here is an issue - the sender won't know which type the actual recipient uses. It could be anything and for most users this is not known at all.

packages/keyring/src/pair/index.spec.ts
```diff
@@ -234,6 +235,51 @@ describe('pair', (): void => {
...
+
+  it('allows encrypt/decrypt with sr25519 keypair', (): void => {
+    const aliceSR25519KeyPair = new Keyring().createFromUri(mnemonicGenerate(), { name: 'sr25519 pair' }, 'sr25519');
+    const bobSR25519KeyPair = new Keyring().createFromUri(mnemonicGenerate(), { name: 'sr25519 pair' }, 'sr25519');
+    const message = new Uint8Array([0x61, 0x62, 0x63, 0x64, 0x65]);
+    const encryptedMessage = aliceSR25519KeyPair.encrypt(message, bobSR25519KeyPair.publicKey);
+
+    expect(bobSR25519KeyPair.decrypt(encryptedMessage)).toEqual(message);
+    expect((): Uint8Array | null => aliceSR25519KeyPair.decrypt(encryptedMessage)).toThrow("Mac values don't match");
+  });
```
jacogr, line 254: We would need an additional test (on the back of the above above comment), where the one party uses ed25519 and the other party using sr25519 and they can exchange messages bi-directionally.

jacogr (commented):
I think this is nice work. However I'm not at all convinced about the requirement to have to specify the key type for the recipient - this is the stickler in making it available to all users, techy and non-techy.

... this is exactly the bigger issue that we are facing with this functionality: multiple key types, without users knowing what they are and them having to magically co-operate.

gdethier:
Hello @jacogr and thank you very much for your review. Sorry to come back to you so late but I had to test a few things. I tried to implement encryption/decryption without any prior knowledge about the key type but was not completely successful.

Here is what I came up with: https://github.com/LaurentTrk/common/pull/1 (in another PR to keep this one "clean" of my experiments). In the code, when no type is provided, the key type is guessed. However, that method does not work in all cases ([this test](https://github.com/LaurentTrk/common/pull/1/files#diff-019aa6ae4ab0432f729b36806dc8ce6c1647fb6719d30fe867b704e327ddf889R265) passes but [this one](https://github.com/LaurentTrk/common/pull/1/files#diff-019aa6ae4ab0432f729b36806dc8ce6c1647fb6719d30fe867b704e327ddf889R275) does not).

I am no cryptographer so I think that I reached the limits of what I can achieve. Suggestions are more than welcome.

packages/keyring/src/keyring.ts
LaurentTrk: Hi @jacogr,
Thanks for your review.
I actually did not catch your point here, the users of this library are not end users, I guess that they are more techy (as  developers) and aware of what key types and encryption are.
I agree that passing the wrong key type for a given public key will fail, but we need this information.

packages/keyring/src/pair/index.spec.ts
LaurentTrk: This test has not been done as you cannot exchange messages if the 2 key types differs.

LaurentTrk:
Hi @gdethier !  
Thanks for your work, that's could be great yes, but unfortunately, sometimes just checking that a key could fit in one or the other type is not enough, some keys fit in boths, but that leads to incorrect encryption/decryption.  
That was the problem in the very initial version of this feature: sr22519 were forced to be treated as ed25519, but that just not work (despite the unit tests poving so, that were using keys that fitted both key types).

jacogr:
Closing, removing this functionality in https://github.com/polkadot-js/common/pull/1762

PR #1350, opened 2022-01-03
Use float64 for u8aEq

jacogr:


jacogr:
This breaks down at the edges - https://github.com/polkadot-js/common/runs/4690260561?check_suite_focus=true#step:3:1579

PR #1352, opened 2022-01-03
Use global `crypto` object for browser and `webcrypto` for Node.js

StefansArya:
This pull request will fix the key pair's signer when using `polkadotApi.Keyring` with `sr25519`.

For referencing about the error:
https://jsbin.com/cewajinaro/edit?html,js,console

```js
let keyring = new polkadotApi.Keyring({
  type: 'sr25519'
});

!async function(){
  await polkadotUtilCrypto.cryptoWaitReady();

  let dummy = window.crypto.getRandomValues(new Uint8Array(32));
  let pair = keyring.addFromSeed(dummy);
  
  let result = pair.sign("Hello"); // throw error: crypto$1.getRandomValues is not a function
  console.log(result);
}();
```

### DIFF
--- a/packages/x-bundle/package.json
+++ b/packages/x-bundle/package.json
@@ -22,6 +22,7 @@
   "dependencies": {
     "@babel/runtime": "^7.16.7",
     "@polkadot/util": "8.2.3-18",
+    "@polkadot/x-global": "8.2.3-18",
     "buffer-es6": "^4.9.3"
   }
 }
--- a/packages/x-bundle/src/crypto.ts
+++ b/packages/x-bundle/src/crypto.ts
@@ -1,4 +1,12 @@
 // Copyright 2017-2022 @polkadot/x-bundle authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-export default {};
+import { xglobal } from '@polkadot/x-global';
+
+let crypto = xglobal.crypto;
+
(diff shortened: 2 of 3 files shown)


jacogr:
Random values are via x-randomvalues

StefansArya:
Hi! Thanks for the response.
`polkadotUtilCrypto.randomAsU8a` also throw similar error for the bundled version of the library.



In my case I can use `window.crypto.getRandomValues` as an alternative for `polkadotUtilCrypto.randomAsU8a(32)`, but sadly `window.crypto.getRandomValues` is also being used when signing transaction data with a **sr25519** key pair. When signing data using **ed25519** or browser extension it doesn't have any problem.

Sorry if it's a bit confusing because I was giving example of the error on the most important feature (signing data) 😅

jacogr:
Ahhh, ok, I just adjusted the test bundle aka https://github.com/polkadot-js/common/blob/master/test-bundle.html to have a `randomAsHex` test and it does fail atm... It sadly tries the `crypto` import, which is invalid.

StefansArya:
Looks great, thanks for pointing out which part that's more appropriate to fix 😁

jacogr:
Closing. Thank you for this, although not merging, it certainly highlighted the issue.

I'm still not sure where this slipped-in, since it was working, so need to really amp up the verification here.

PR #1354, opened 2022-01-04
Use x-global inside x-randomvalues

jacogr:
Replacement for https://github.com/polkadot-js/common/pull/1352

(With local test-bundle.html verification)

### DIFF
--- a/packages/x-randomvalues/src/browser.ts
+++ b/packages/x-randomvalues/src/browser.ts
@@ -1,8 +1,13 @@
 // Copyright 2017-2022 @polkadot/x-randomvalues authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import { xglobal } from '@polkadot/x-global';
+
 export { packageInfo } from './packageInfo';
 
 export function getRandomValues <T extends Uint8Array> (arr: T): T {
-  return crypto.getRandomValues(arr);
+  // We use x-global here - this prevents packagers such as rollup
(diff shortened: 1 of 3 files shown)


packages/x-randomvalues/src/browser.ts
```diff
@@ -1,8 +1,13 @@
 // Copyright 2017-2022 @polkadot/x-randomvalues authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
+import { xglobal } from '@polkadot/x-global';
+
 export { packageInfo } from './packageInfo';
```
StefansArya, line 7: It seems works for Browser and Deno.
But, can we just add this line instead?

```suggestion
  let crypto = xglobal.crypto || require('crypto').webcrypto;
```

It will also support for Node.js if the developer prefer to import the bundled version. But I think it's low priority for now, and it may only happen if Node.js users is importing through [HTTPS module loader](https://nodejs.org/api/esm.html#https-loader) instead of installing through package manager.

jacogr: The bundled is not for node. If we wish to support node, it would mean adding an `x-crypto` and using export maps - we don't use inline requires anywhere.

StefansArya: Hmm, ok, let's forget about it as using package manager is already a working solution for Node.js.

PR #1436, opened 2022-03-18
Update genesis hash of Dock mainnet

lovesh:
Signed-off-by: lovesh <lovesh.bond@gmail.com>

### DIFF
--- a/packages/networks/src/defaults/genesis.ts
+++ b/packages/networks/src/defaults/genesis.ts
@@ -24,6 +24,7 @@ export const knownGenesis: KnownGenesis = {
     '0x67dddf2673b69e5f875f6f25277495834398eafd67f492e09f3f3345e003d1b5'
   ],
   'dock-mainnet': [
+    '0x6bfe24dca2a3be10f22212678ac13a6446ec764103c0f3471c71609eac384aae',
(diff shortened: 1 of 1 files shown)


packages/networks/src/defaults/genesis.ts
```diff
@@ -24,7 +24,7 @@ export const knownGenesis: KnownGenesis = {
     '0x67dddf2673b69e5f875f6f25277495834398eafd67f492e09f3f3345e003d1b5'
   ],
   'dock-mainnet': [
-    '0xf73467c6544aa68df2ee546b135f955c46b90fa627e9b5d7935f41061bb8a5a9'
+    '0x6bfe24dca2a3be10f22212678ac13a6446ec764103c0f3471c71609eac384aae'
```
jacogr, line 27: You really want to keep the old and new entries - this allows accounts to be migrated. (Newest first, see the top-level comment and the Kusama example)

Without this accounts will be "lost" that have been previously assigned to the network.

jacogr (changes requested):

packages/networks/src/defaults/genesis.ts
lovesh: Thanks for the feedback :). Updated.

jacogr (approved):

PR #1467, opened 2022-04-27
feat(networks): add statemint to KnownLedger

carlosala:
We (Zondax) are developing the Statemint app (as we developed also the Statemine one) and we'll need to use it in polkadot.js.org/apps.
If there's something wrong, please tell me and I'll try to answer you as soon as possible!

### DIFF
--- a/packages/networks/src/defaults/ledger.ts
+++ b/packages/networks/src/defaults/ledger.ts
@@ -20,5 +20,6 @@ export const knownLedger: KnownLedger = {
   polymesh: 0x00000253,
   sora: 0x00000269,
   statemine: 0x000001b2, // common-good on Kusama, shares derivation
+  statemint: 0x00000162, // common-good on Polkadot, shares derivation
   xxnetwork: 0x000007a3
 };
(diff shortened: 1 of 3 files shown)


packages/networks/src/defaults/ledger.ts
```diff
@@ -20,5 +20,6 @@ export const knownLedger: KnownLedger = {
   polymesh: 0x00000253,
   sora: 0x00000269,
   statemine: 0x000001b2, // common-good on Kusama, shares derivation
+  statemint: 0x00000162, // common-good on Polkadot, shares derivation
```
jacogr, line 23: As per the comment above that explains the additions here, additional information is required -

- The genesis has for the network should be added to `genesis.ts`
- The Ledger app should be added to `hw-ledger/defaults.ts`

Will approve the CI run (since it is a first contribution), but since it does not have the above, it will fail the checks.

jacogr (dismissed):

packages/networks/src/defaults/ledger.ts
carlosala: Yes, I saw that later. I'm fixing it. Thanks for the fast answer!

jacogr (approved):
Thank you.

PR #1490, opened 2022-05-16
add parallel to KnownLedger

carlosala:
We (at Zondax) are developing the [Parallel](https://parallel.fi) Ledger app.


jacogr (approved):

carlosala:
Hi @jacogr!
By the way, thanks for merging that fast! We would like to begin testing the parallel app as soon as possible, and it would be really helpful if you release a new version of common packages. I have also tried with `yarn link` but yarn is lost with monorepos 😅 
Thanks!

PR #1497, opened 2022-05-24
some changes to KnownLedger

carlosala:
I list the changes here:
- Nodle has changed chain, now with a new parachain. I added it but I left the other one, just for compatibility some time. It should be removed in some time. EDIT: I asked the team and they confirmed that the old chain has been halted, so I removed it from those files.
- I added Astar and Polkadex to KnownLedger

I just have one doubt. The name of the keys (so `astar`, `nodle-para`, `nodle-chain`) is the name that I can see in polkadot.js.org/apps at the left side, just below the name, right? So `<name>/<spec version>`? In this case polkadex name is `node` that probably is so general, but I added a comment to recognize it.
Thanks!

### DIFF
--- a/packages/hw-ledger/src/defaults.ts
+++ b/packages/hw-ledger/src/defaults.ts
@@ -18,6 +18,7 @@ export const ledgerApps: Record<string, (transport: Transport) => SubstrateApp> 
   karura: newKaruraApp,
   kusama: newKusamaApp,
   'nodle-chain': newNodleApp,
+  'nodle-para': newNodleApp,
   parallel: newParallelApp,
   polkadot: newPolkadotApp,
   polymesh: newPolymeshApp,
--- a/packages/hw-ledger/src/defaults.ts
+++ b/packages/hw-ledger/src/defaults.ts
@@ -4,11 +4,12 @@
 import type Transport from '@ledgerhq/hw-transport';
 import type { SubstrateApp } from '@zondax/ledger-substrate';
 
-import { newAcalaApp, newBifrostApp, newCentrifugeApp, newDockApp, newEdgewareApp, newEquilibriumApp, newGenshiroApp, newKaruraApp, newKusamaApp, newNodleApp, newParallelApp, newPolkadotApp, newPolymeshApp, newSoraApp, newStatemineApp, newStatemintApp, newXXNetworkApp } from '@zondax/ledger-substrate';
+import { newAcalaApp, newAstarApp, newBifrostApp, newCentrifugeApp, newDockApp, newEdgewareApp, newEquilibriumApp, newGenshiroApp, newKaruraApp, newKusamaApp, newNodleApp, newParallelApp, newPolkadexApp, newPolkadotApp, newPolymeshApp, newSoraApp, newStatemineApp, newStatemintApp, newXXNetworkApp } from '@zondax/ledger-substrate';
 
 // These match up with the keys of the knownLedger object in the @polkadot/networks/defaults/ledger.ts
 export const ledgerApps: Record<string, (transport: Transport) => SubstrateApp> = {
   acala: newAcalaApp,
+  astar: newAstarApp,
   bifrost: newBifrostApp,
   centrifuge: newCentrifugeApp,
   'dock-mainnet': newDockApp,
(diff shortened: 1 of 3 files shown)


packages/hw-ledger/src/defaults.ts
```diff
@@ -17,7 +18,8 @@ export const ledgerApps: Record<string, (transport: Transport) => SubstrateApp>
   genshiro: newGenshiroApp,
   karura: newKaruraApp,
   kusama: newKusamaApp,
-  'nodle-chain': newNodleApp,
+  node: newPolkadexApp,
```
jacogr, line 21: `node` is actually the Substrate default, so should not be used. 

In this case however, the mapping internally here is only done on the key (i.e. key needs to be the same for app, Ledger version & genesis), however to determine which chain to use, the mapping is done on genesis.

So it _should_ be 100% fine to change this to `polkadex` and it it doesn't quite do what is expected, the user of the lib should map on genesis, not name.

packages/networks/src/defaults/genesis.ts
```diff
@@ -53,8 +53,11 @@ export const knownGenesis: KnownGenesis = {
     '0xe3777fa922cafbff200cadeaea1a76bd7898ad5b89f7848999058b50e715f636', // Kusama CC2
     '0x3fd7b9eb6a00376e5be61f01abb429ffb0b104be05eaff4d458da48fcd425baf' // Kusama CC1
   ],
-  'nodle-chain': [
-    '0xa3d114c2b8d0627c1aa9b134eafcf7d05ca561fdc19fb388bb9457f81809fb23'
+  node: [ // This is Polkadex Mainnet
```
jacogr, line 56:
```suggestion
  polkadex: [
```

Lint will probably complain about ordering after this

packages/hw-ledger/src/defaults.ts
jacogr, line 21:
```suggestion
  polkadex: newPolkadexApp,
```

packages/networks/src/defaults/ledger.ts
```diff
@@ -17,7 +18,8 @@ export const knownLedger: KnownLedger = {
   genshiro: 0x05f5e0fc,
   karura: 0x000002ae,
   kusama: 0x000001b2,
-  'nodle-chain': 0x000003eb,
+  node: 0x0000031f, // This is Polkadex Mainnet
```
jacogr, line 21:
```suggestion
  polkadex: 0x0000031f,
```

jacogr:
As per the genesishash/name comment above -

- this is how the apps UI uses it - https://github.com/polkadot-js/apps/blob/e9bb234b1fa9f4d7f227808fc0c9ff6a00db0b40/packages/react-hooks/src/useLedger.ts#L60 (via genesis)
- this is how the extension uses it - https://github.com/polkadot-js/extension/blob/885a67e59624f0da4fa02ab5e8d6a3327c16ef92/packages/extension-ui/src/hooks/useLedger.ts#L49-L51 (via genesis)

I'm hoping that most consumers would follow the same approach (at least the extension would be reasonably safe, most of them are building on what is linked above)

packages/hw-ledger/src/defaults.ts
carlosala: Thanks Jaco! By the way, if something is not working I'll talk to Polkadex team to change the node name. But it makes totally sense!

carlosala: I'll push the changes

carlosala:
> As per the genesishash/name comment above -
> 
> * this is how the apps UI uses it - [polkadot-js/apps@`e9bb234`/packages/react-hooks/src/useLedger.ts#L60](https://github.com/polkadot-js/apps/blob/e9bb234b1fa9f4d7f227808fc0c9ff6a00db0b40/packages/react-hooks/src/useLedger.ts#L60) (via genesis)
> * this is how the extension uses it - [polkadot-js/extension@`885a67e`/packages/extension-ui/src/hooks/useLedger.ts#L49-L51](https://github.com/polkadot-js/extension/blob/885a67e59624f0da4fa02ab5e8d6a3327c16ef92/packages/extension-ui/src/hooks/useLedger.ts#L49-L51) (via genesis)
> 
> I'm hoping that most consumers would follow the same approach (at least the extension would be reasonably safe, most of them are building on what is linked above)

Thanks for the clarify!

PR #1515, opened 2022-06-05
add Composable support for ledger

carlosala:
Hi!
I also changed the genesis hash for Centrifuge, as they migrated to a new chain.
Thanks Jaco!

### DIFF
--- a/packages/networks/src/defaults/genesis.ts
+++ b/packages/networks/src/defaults/genesis.ts
@@ -21,7 +21,8 @@ export const knownGenesis: KnownGenesis = {
     '0x9f28c6a68e0fc9646eff64935684f6eeeece527e37bbe1f213d22caa1d9d6bed'
   ],
   centrifuge: [
-    '0xb3db41421702df9a7fcac62b53ffeac85f7853cc4e689e0b93aeb3db18c09d82'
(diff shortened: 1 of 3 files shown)


packages/networks/src/defaults/genesis.ts
```diff
@@ -21,7 +21,10 @@ export const knownGenesis: KnownGenesis = {
     '0x9f28c6a68e0fc9646eff64935684f6eeeece527e37bbe1f213d22caa1d9d6bed'
   ],
   centrifuge: [
-    '0x67dddf2673b69e5f875f6f25277495834398eafd67f492e09f3f3345e003d1b5'
+    '0xb3db41421702df9a7fcac62b53ffeac85f7853cc4e689e0b93aeb3db18c09d82'
```
jacogr, line 24: Would people expect their old tied accounts to be available on this network? In that case, the new genesis should just be added as the first entry.

carlosala: The new chain uses the same derivation path and prefix as the old chain, so if they add an account using ledger they are going to retrieve the same account. By the way, we can just append it to allow people using their accounts without readding them.

jacogr (approved):
As always, thank you!

carlosala:
Thank you too!☺️😉

PR #1641, opened 2022-09-06
feat(networks): add phala and khala to knownLedger

carlosala:
Hi Jaco!
Let's add two new chains.
Thanks for all your work!

PS: with the new implementation is far easier than before, good job there! 😉


jacogr:
Phew :) Happy that it is easier for you. Not perfect as of yet, but seems to align closer with the base ZondaX libs, which is great.

jacogr (approved):

carlosala:
Thanks @jacogr.
Just one more thing, there was a typo in our package, could you please update it to `v0.36.1`, so latest one?
Thanks! 😉

PR #1689, opened 2022-12-06
add French mnemonic file

poka-IT:


jacogr:
This will actually not get used. The reasons being -

1. The [WASM bip39 execution path](https://github.com/polkadot-js/wasm/blob/4ceb620462478db9682e7b4512c37ae8f8e885b9/packages/wasm-crypto/src/rs/bip39.rs#L16-L22) is the one that gets executed in 99.9% of the cases (JS is only used as a fallback when WASM is not available in the environment)
2. The actual JS code here only used the `DEFAULT_WORDLIST` import (I believe CI will actually complain about the unused import here)

So basically with this is won't do anything. There is a way to do multiple languages, but it won't be simple.

1. The WASM environment needs to take a wordlist as input (the bonus is that this reduces bundle size, i.e. currently both the WASM and JS environment has a copy)
2. Execution needs to use this list
3. The same logic for specification should happen in the JS environment (it _could_ be as simple as `setWordlist(...)` before execution happens)

Overall, since each list adds to the bundle sizes (which is a huge ongoing problem), it will therefore need to be up to the developer using it to specify, instead of "having all available always".

poka-IT:
I know it's a big job.
Do you have plans (I mean the team) to do this soon ?
Just to know

jacogr:
The team is basically me for anything and everything in polkadot-js. It is certainly something I _want_ to do (I did get the Rust bip39 dev to make the required changes to his lib), but it has only been hovering with no action taken. The reality is that most probably unless there is a PR that comes in with all the moving parts, it will take "forever" to get to.

(It is also not even logged, i.e. it only exists in my min, see https://github.com/polkadot-js/wasm/issues)

jacogr:
The way to introduce this has been logged https://github.com/polkadot-js/wasm/issues/452 - once available we can have this functionality.

PR #1699, opened 2023-01-06
fix Nodle's known network name

aliXsed:
Nodle's known network name in [ss58-registry.json](https://github.com/paritytech/ss58-registry/blob/main/ss58-registry.json) is "nodle". In contrast "nodle-para" is the runtime's spec name shared between nodle's main parachain on Polkadot and nodle's test parachain on Rococo. This change is necessary for the logic in [networks/src/interfaces.ts](https://github.com/polkadot-js/common/blob/master/packages/networks/src/interfaces.ts) to flag Nodle as a selectable network to interact with Ledger. This issue is impacting polkadot-js/apps in not listing Nodle in the modal dialog for creating a new Ledger account.
This fix needs to be checked by @carlosala in case it may have any side-effect on the actual function of connecting to Nodle's app on Ledger.

### DIFF
--- a/packages/hw-ledger/src/defaults.ts
+++ b/packages/hw-ledger/src/defaults.ts
@@ -22,7 +22,7 @@ export const ledgerApps: Record<string, string> = {
   karura: 'Karura',
   khala: 'Khala',
   kusama: 'Kusama',
-  'nodle-para': 'Nodle',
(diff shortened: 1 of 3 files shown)


carlosala (approved):
LGTM👌🏻

jacogr (approved):

PR #1722, opened 2023-02-04
feat: allow setting of thousand/decimal separators

### LINKED ISSUE #636: Allow setting of thousand/decimal seperators
formatBalance & formatNumber

alko89:
Been trying to get an issue to get started with some development on polkadot and found this and gave it a go. 

Sets thousand and decimal separators based on locale.

Fix #636


packages/util/src/number/getSeparator.spec.ts
```diff
@@ -0,0 +1,72 @@
+// Copyright 2017-2023 @polkadot/util authors & contributors
```
jacogr, line 1: I would keep this file in `format/*` and not export it from index. It really is specific to that functionality.

packages/util/src/number/getSeparator.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2017-2023 @polkadot/util authors & contributors
```
jacogr, line 1: As above with getSeperator.spec.ts - keep in `format/*`

packages/util/src/number/index.ts
```diff
@@ -5,5 +5,6 @@
  * @summary Utility methods to convert to and from `number` values
  */
 
+export { getSeparator } from './getSeparator';
```
jacogr, line 8: Just remove the export after the previous moves.

jacogr (commented):
Thank you. Think this is sensible and you did the right thing keeping the existing behavior and making the new opt-in. 

(My wife, which is German, is actually driven slightly batty by the `.` and `,` in the "wrong" positions on the apps UI, so having a setting there at some point to "default" (as-is) or "detected locale/specific locale" would at least make my life happier :))

alko89:
Hehe, yeah we have separators inverted in Slovenia as well, I prefer to use en locale on my OS because I rather use the format that is more common on the internet. I'm annoyed by my bank website that forces sl locale on their site and have to be careful that I don't accidentally send someone 12.30 (1230 €) instead of 12,30 (12.3 €) :sweat_smile: 

Its annoying this is even a thing, but at least its not as big of a can of worms as...dates!

packages/util/src/number/getSeparator.spec.ts
alko89: Done, check

jacogr (approved):
Thank you!

PR #1725, opened 2023-02-07
feat(networks): support ledger for picasso

carlosala:
Hi Jaco!
Another Ledger app. Could you please release here and in @polkadot/apps? The team is a bit in a rush with the testing and all the stuff 😅 
Thanks!


jacogr:
A release can only happen over a weekend - there are a lot of stuff that pull in this that all needs a release when bumped. (All this has a knock-on ecosystem effect).

I am probably going to bump all the common deps over the coming weekend along with the other releases - at least that is the plan currently.

jacogr (approved):

PR #1840, opened 2023-06-05
feat(networks): add zeitgeist Ledger support

carlosala:
I forgot to add this app in the past, here you have!
Thanks Jaco :)


jacogr:
We just need to adjust this test - https://github.com/polkadot-js/common/actions/runs/5175216722/jobs/9322527701?pr=1840#step:5:85

(There was a reason for checking in the past, although currently it seems like a pain - although one that doesn't change often. So just adjust that test as well to take the new lowest value into account)

jacogr (approved):

PR #1855, opened 2023-07-16
add Westmint genesis hash

Nick-1979:


jacogr:
is there a specific reason for this?

For the most part (with the exception of the test relays), we have not been adding any dev/test genesisHashes. The reason being that they do change a lot and is a pain to keep in-check. Testnets are too volatile and meant to "go dead" or "be reset", which has a longevity issue.

With the addition here, it also misses any previous hashes (if they do/don't exist, have not kept track of any test resets here at all - specifically with relay resets, these do change), which is needed.

Nick-1979:
Yes, you are correct. However, as we aim for users to become acquainted with the functionality of these test chains, we consistently include them in the packages manually. Adding them to the package streamlines our workflow.

jacogr:
Closing.

The best approach is to keep track of these yourself, since there is no process at all to track these inside polkadot-js. (Apart from the relays which do get monitored). 

As with everything, adding it is one thing, maintaining it indefinitely is a different kettle-of-fish - and in this case the overhead that is associated with that (completely manual), including the "no testnets" setup just adds too much. (I'll check, if not made explicit already, will add a prominent note to not allow testnets to this file)

PR #1866, opened 2023-08-21
Fix decoding i8/i16 by using DataView and preventing future false use of `u8aToNumber` by deprecating it

peetzweg:
change decoding of byte to number methods for numbers <=32 bits, furthermore limiting to use i8/16/32

using DataView api do implement simpler decoding of these fixed width numbers. Thereby deprecating `u8aToNumber()`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView

Needed for https://github.com/polkadot-js/api/issues/5702 and https://github.com/polkadot-js/api/pull/5703

See https://github.com/polkadot-js/api/pull/5703#issuecomment-1684059771 for what was wrong with the previous implementation.

PR #1867, opened 2023-08-21
Check sign flag for i{8, 16, 32, ...} conversions

jacogr:
Replaces https://github.com/polkadot-js/common/pull/1866


packages/util/src/u8a/toNumber.ts
```diff
@@ -15,7 +15,11 @@ interface ToNumberOptions {
 export function u8aToNumber (value: Uint8Array, { isNegative = false }: ToNumberOptions = {}): number {
   const count = value.length;
 
-  if (isNegative) {
+  // When the value is a i{8, 16, 24, 32, 40, 40} values and the top-most bit
+  // indicates a signed value, we use a two's complement conversion. If one of these
+  // flags are not set, we just do a normal unsigned conversion (the same shortcut
+  // applies in both the u8aTo{BigInt, Bn} conversions as well)
+  if (isNegative && (value[count - 1] & 0x80)) {
```
peetzweg, line 22: Is it fine if this goes out of bound?
Also wouldn't we need to check the overall first bit, so basically always `value[0]`?

peetzweg: I did this in my branch once as well, keep in mind we need to make a distinction between `positive/negative`, yes. But also between `signed/unsigned` in the first place as well.

We need one more flag or two distinct function for each case.

packages/util/src/u8a/toNumber.spec.ts
```diff
@@ -3,114 +3,29 @@
...
-          { isNegative: true }
-        )
-      ).toBe(-9999999999999);
-    });
+describe('u8aToNumber', (): void => {
+  describe('conversion tests', (): void => {
+    for (let i = 0, count = TESTS_NUM.length; i < count; i++) {
+      const [, isNegative, numarr, strval] = TESTS_NUM[i];
+
+      it(`${i}: creates ${strval} (bitLength=${numarr.length * 8}, isNegative=${isNegative})`, (): void => {
```
peetzweg, line 20: Smart, we have the power of js in tests as well. 🤦

packages/util/src/u8a/toBigInt.ts
```diff
@@ -23,7 +23,7 @@ export function u8aToBigInt (value: Uint8Array, { isLe = true, isNegative = fals
     : value.slice().reverse();
   const count = u8a.length;
 
-  if (isNegative && (!count || (u8a[count - 1] & 0x80))) {
+  if (isNegative && count && (u8a[count - 1] & 0x80)) {
```
peetzweg, line 26: So to understand this, `isNegative` will not indicate if we want to convert a `signed` value, correct?
If so maybe it would be good to refactor the name to `isSigned` and deprecate `isNegative` option?

jacogr: The naming `isNegative` is indeed very unfortunate, it does mean `isSigned` and should have been named such 5 year ago. Got it right eventually [in the api](https://github.com/polkadot-js/api/blob/52e36534cfde94d4f032e88fee9ba0d24ccb4fb1/packages/types-codec/src/abstract/Int.ts#L97).

The timelines for deprecation (at least 6 months and then in the next major) and then the additional clutches it brings code & maintenance wise to support old/new has newer seemed to be worth the effort. So it is a wart.

PR #1878, opened 2023-09-06
feat: `isRiscV` guard function 

peetzweg:
Introduced in originally in this PR https://github.com/polkadot-js/api/pull/5699#pullrequestreview-1604611089 , it was suggested to be moved to the `@polkadot/common` package. 

I've basically just copied over the code form the pr in the api package. The `isWasm` and `isRiscV` function signature differs as the `isWasm`'s parameter of bytes is optional. Should we align both functions to have a optional or required parameter? 

https://github.com/peetzweg/common/blob/d8a4d018f18fe69f2ad796d986ff950335cb2b35/packages/util/src/is/riscv.ts#L15-L17

https://github.com/peetzweg/common/blob/d8a4d018f18fe69f2ad796d986ff950335cb2b35/packages/util/src/is/wasm.ts#L15-L17

### DIFF
--- a/packages/util/src/is/index.ts
+++ b/packages/util/src/is/index.ts
@@ -25,6 +25,7 @@ export { isNumber } from './number.js';
 export { isObject } from './object.js';
 export { isObservable } from './observable.js';
 export { isPromise } from './promise.js';
+export { isRiscV } from './riscv.js';
 export { isString } from './string.js';
 export { isTestChain } from './testChain.js';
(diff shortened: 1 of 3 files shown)


jacogr:
I think it is fine as-is - `unknown` also caters for `undefined`. (So `isWasm` seems to go a bit overboard in this case - I think this is the case with a lot of functions, where `?` is not strictly required...)

jacogr (approved):
Thank you.

PR #1928, opened 2024-07-12
Add peaq support for ledger

lavish0000:
add ledger support for peaq

peaq is listed in - 

BIP-0044 - https://github.com/satoshilabs/slips/blob/master/slip-0044.md
ledger-substrate-js - https://github.com/Zondax/ledger-substrate-js/pull/115

### DIFF
--- a/packages/hw-ledger/package.json
+++ b/packages/hw-ledger/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@polkadot/hw-ledger-transports": "13.0.1",
     "@polkadot/util": "13.0.1",
-    "@zondax/ledger-substrate": "0.44.4",
+    "@zondax/ledger-substrate": "0.44.7",
     "tslib": "^2.6.2"
   }
 }
--- a/packages/hw-ledger/src/defaults.ts
+++ b/packages/hw-ledger/src/defaults.ts
@@ -27,6 +27,7 @@ export const ledgerApps: Record<string, string> = {
   nodle: 'Nodle',
   origintrail: 'OriginTrail',
   parallel: 'Parallel',
+  peaq: 'Peaq',
   pendulum: 'Pendulum',
   phala: 'Phala',
   picasso: 'Picasso',
--- a/packages/networks/src/defaults/genesis.ts
+++ b/packages/networks/src/defaults/genesis.ts
@@ -104,6 +104,9 @@ export const knownGenesis: KnownGenesis = {
   parallel: [
     '0xe61a41c53f5dcd0beb09df93b34402aada44cb05117b71059cce40a2723a4e97'
   ],
+  peaq: [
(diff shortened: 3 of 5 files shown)


lavish0000:
@jacogr , Can you please review this?

TarikGul (approved):

TarikGul:
This will be out in 13.0.2 shortly

lavish0000:
Thank you so much, @TarikGul.

I have another question: When will the ledger icon appear on the Polkadot.js app for peaq or within our network in the Polkadot.js extension list of ledger devices?

Additionally, do we need to make any changes in other Polkadot.js repositories to add ledger support?

TarikGul:
> Additionally, do we need to make any changes in other Polkadot.js repositories to add ledger support?

As long as you have a running chain is Polkadot.js this is all that is required (I am almost positive, and will let you know otherwise).

> When will the ledger icon appear on the Polkadot.js app for peaq or in the Polkadot.js extension list of networks that support ledger devices?

For apps, and the extension it should be the same but its based off of if it identifies the chain as `havingLedgerSupport` which is enabled based on the configs you passed in in this PR. So you should be alright.

TarikGul:
Ledger in apps is going to be fixed first this week with the new updates and then the extension. We have to update the interface for how the extension interacts with dApps since it now requires retrieving the metadata blob for constructing the proof to attach to the tx for signing. So you will see incoming PR's to have all this fixed soon.

lavish0000:
Thanks for clarifying, @TarikGul 

I'm asking because I'm unable to see the ledger icon in Polkadot.js apps and extensions. I ran the Polkadot.js apps locally and found that the issue is due to "@polkadot/networks" needing to be updated to version 13.0.2 for it to work. 
So basically dependency update is needed.

Do you know how dependency update is usually handled? Does the Polkadot team update it periodically, or do I need to create a pull request (PR) for this?

TarikGul:
> Thanks for clarifying, @TarikGul
> 
> I'm asking because I'm unable to see the ledger icon in Polkadot.js apps and extensions. I ran the Polkadot.js apps locally and found that the issue is due to "@polkadot/networks" needing to be updated to version 13.0.2 for it to work. So basically dependency update is needed.
> 
> Do you know how dependency update is usually handled? Does the Polkadot team update it periodically, or do I need to create a pull request (PR) for this?

I am actually the one who does all the downstream updates. Everything but apps will be updated by Monday then we will update the common deps in apps during the week, but it also requires a large rework of the Ledger interface in apps, so it wont be immediate. 

The release Process is not as straightforward as updating the deps in one place unfortunately everything has layers in PJS - meaning things have to be updated in there tree order: wasm -> common -> api -> phishing -> tools -> ui -> extension -> apps

So it takes a few days to get everything updated tested etc. And in this case because 13.0.2 is going to be a huge change for apps it will take an extra day or so.

PR #1941, opened 2024-09-12
Add bittensor to genericLedgerApps

camfairchild:
Bittensor has applied the runtime upgrade for the generic Ledger app integration.
This PR adds it to the mapping of `genericLedgerApps`


garrett-opentensor (approved):

TarikGul (changes requested):
You will also need to add configuration to 

-`packages/networks/src/defaults/genesis.ts`
-`packages/networks/src/defaults/ledger.ts`

camfairchild:
> You will also need to add configuration to
> 
> -`packages/networks/src/defaults/genesis.ts` -`packages/networks/src/defaults/ledger.ts`

Apologies. I've added the network to `packages/networks/src/defaults/ledger.ts` as it's already in the `genesis.ts`

TarikGul (approved):

PR #1973, opened 2025-01-31
Fix Misclassification of ECDSA Signatures in `verifyMultisig()`

### LINKED ISSUE #1898: Signing and verifying a message using ECDSA sometimes fails
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

**Expected behaviour:** I should be able to generate a KeyringPair that signs messages using the ECDSA algorithm and verify the signatures against the message + the KeyringPair's public key

**Current behaviour:** This works *most* of the time, approx 99/100 times, but sometimes the verification fails.

* **What is the motivation for changing the behavior?**

I'm working on a project that relies on being able to sign and verify data from the KeyringPair or from Polkadot wallets so I want to be able to rely on this functionality.

* **Please tell us about your environment:**

I'm using `@polkadot/keyring`, `@polkadot/util` and `@polkadot/util-crypto` at the latest version as of writing (12.6.1).

I've created a minimal repository here that reproduces the bug: https://github.com/rjwebb/polkadot-ecdsa-bug-reproduction

  - Version: 12.6.1
  - Environment: Node v18.17.1

    - [x] Node.js
    - [ ] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
    - [x] TypeScript (include tsc --version) 5.2.2
    - [ ] Other

valentinfernandez1:
This PR addresses issue #1898 by refining the logic in `verifyMultisig()` to correctly distinguish between `ECDSA`, `Ed25519`, and `Sr25519` signatures.

**Problem:**
Previously, `verifyMultisig()` misidentified certain `ECDSA` signatures (65 bytes) as `Ed25519` or `Sr25519` signatures (64 bytes) when the first byte of the `ECDSA` signature was `0` or `1`. This happened because the function interpreted the first byte as a multisignature prefix rather than part of the `ECDSA` signature itself.

**Solution:**
- Improved signature length handling:
  - If the signature is 66 bytes, it is treated as an `ECDSA` signature prefixed with a multisiginature prefix byte, so the prefix is removed before verification.
  - If the signature is 65 bytes, it could be:
    - A proper `ECDSA` signature, which should remain unchanged for verification.
    - An `Ed25519`/`Sr25519` signature (1-byte prefix + 64-byte signature), in which case the prefix is removed before verification.
- Improved verification order (when the signature starts with `0` or `1` and 65 bytes of lenght):
  - First, attempt `Ed25519`/`Sr25519` verification by removing the prefix.
  - If verification fails, fall back to `ECDSA` verification using the full 65-byte signature.
  
  closes #1898


TarikGul (approved):
🚀

PR #1974, opened 2025-02-04
Enable KeyringPair Creation with Custom Mnemonic Wordlists

### LINKED ISSUE #1804: Add support for bip39 with other wordlists
We are pulling in the extra wordlists via https://github.com/polkadot-js/common/pull/1803

This has been PR-ed previously in https://github.com/polkadot-js/common/pull/1689 and there has been some discussion. However after giving it some thought, we could actually make this work.

### part 1

1. We need a `wordlist` option passed to the bip39 functions, e.g. last param of `wordlist: string[] = DEFAULT_WORDLIST` - 

https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/bip39.ts#L53
https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/bip39.ts#L95

2. We change the `DEAFULT_WORDLIST` usage in these functions so be `wordlist` instead.
3. Add actual tests for these in `bip39.spec.ts` to ensure we are actually ok.

### part 2

Now the base generation is ok, but we may still use the wasm versions which doesn't have support for this. To get around it -

1. Adapt the wasm/js check inside the exposed functions to also take in a last param of `wordlist?: string[], onlyJs?: boolean`

https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/validate.ts#L22-L23
https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/toMiniSecret.ts#L11
https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/toLegacySeed.ts#L27
https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/toEntropy.ts#L9
https://github.com/polkadot-js/common/blob/91a2eac220df188a5c339f2dcc6ca98e5896da0b/packages/util-crypto/src/mnemonic/generate.ts#L21

2. For these treat the `wordlist` the same as `onlyJs`, i.e. when passed in we want to take the JS (non-WASM) route, i.e. it ties into the changes we made in `bip39.ts`. This looks something like `!hasBigInt || (!wordlist && !onlyJs && isReady())` and the param passed on the JS-only path
3. For all the above add tests...

### part 3

Now we also need keyring support -

1. Make the required wordlist-in changes to `addFromMnemonic`, `addFromUri` & `createFromUri` in https://github.com/polkadot-js/common/blob/892004563faa65dcd81bf60cf8d2e5aba74ae9d4/packages/keyring/src/keyring.ts#L201-L223
2. Add tests...

valentinfernandez1:
This pull request addresses part 3 of issue #1804.  It introduces the capability to create `KeyringPair` instances using custom mnemonic wordlists, enhancing support for various languages and improving flexibility in key pair generation.

Key Changes:
  - Added functionality to specify custom BIP-39 mnemonic wordlists during KeyringPair creation using `addFromMnemonic()`, `addFromUri()` and `createFromUri()`.
  - Updated relevant documentation to reflect the new feature.
  - Included tests to ensure compatibility with different mnemonic wordlists.

closes #1804

### DIFF
--- a/packages/keyring/src/index.spec.ts
+++ b/packages/keyring/src/index.spec.ts
@@ -6,7 +6,8 @@
 import type { KeyringPair$Json } from './types.js';
 
 import { hexToU8a, stringToU8a } from '@polkadot/util';
-import { base64Decode, cryptoWaitReady, encodeAddress, randomAsU8a, setSS58Format } from '@polkadot/util-crypto';
+import { base64Decode, cryptoWaitReady, encodeAddress, mnemonicGenerate, randomAsU8a, setSS58Format } from '@polkadot/util-crypto';
+import * as languages from '@polkadot/util-crypto/mnemonic/wordlists/index';
 
 import { decodePair } from './pair/decode.js';
 import Keyring from './index.js';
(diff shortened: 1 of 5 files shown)


TarikGul (approved):
Amazing, nice job 🚀

PR #1978, opened 2025-02-17
Cast ws as global.WebSocket

valentinfernandez1:
Fixes an issue where the `WebSocket` type was inconsistently resolved as either the ws library's WebSocket implementation or the native `WebSocket`, leading to type mismatches.


TarikGul (approved):
👍

PR #1983, opened 2025-04-13
Set pbkdf2Encode rounds to default to 210,000

TarikGul:
In line with https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2 - set the default rounds to 210_000.

## Why 210,000?:

We pass in `Sha512` for `pbkdf2Js` from `@noble/hashes`
By default `pbkdf2` from `@polkadot/wasm-crypto` uses `HMAC-SHA512`.

### DIFF
--- a/packages/util-crypto/src/pbkdf2/encode.ts
+++ b/packages/util-crypto/src/pbkdf2/encode.ts
@@ -15,7 +15,7 @@ interface Result {
   salt: Uint8Array;
 }
 
-export function pbkdf2Encode (passphrase?: string | Uint8Array, salt: Uint8Array = randomAsU8a(), rounds = 2048, onlyJs?: boolean): Result {
+export function pbkdf2Encode (passphrase?: string | Uint8Array, salt: Uint8Array = randomAsU8a(), rounds = 210000, onlyJs?: boolean): Result {
   const u8aPass = u8aToU8a(passphrase);
   const u8aSalt = u8aToU8a(salt);
 


packages/util-crypto/src/hd/ledger/master.ts
```diff
@@ -9,8 +9,8 @@ import { mnemonicToSeedSync } from '../../mnemonic/bip39.js';
 const ED25519_CRYPTO = 'ed25519 seed';
 
 // gets an xprv from a mnemonic
-export function ledgerMaster (mnemonic: string, password?: string): Uint8Array {
-  const seed = mnemonicToSeedSync(mnemonic, password);
+export function ledgerMaster (mnemonic: string, password?: string, rounds?: number): Uint8Array {
```
ap211unitech, line 12: Can we add some docs here explaining what `rounds` means?

packages/util-crypto/src/mnemonic/toMiniSecret.ts
```diff
@@ -8,7 +8,7 @@ import { pbkdf2Encode } from '../pbkdf2/index.js';
 import { mnemonicToEntropy } from './toEntropy.js';
 import { mnemonicValidate } from './validate.js';
 
-export function mnemonicToMiniSecret (mnemonic: string, password = '', wordlist?: string[], onlyJs?: boolean): Uint8Array {
+export function mnemonicToMiniSecret (mnemonic: string, password = '', wordlist?: string[], onlyJs?: boolean, rounds?: number): Uint8Array {
```
ap211unitech, line 11: Here as well

ap211unitech (changes requested):

ap211unitech (approved):

PR #1990, opened 2025-06-09
Add input length validation in `scryptFromU8a` function

ap211unitech:
## 📝 Description

This PR addresses unsafe parsing in `scryptFromU8a` (`packages/util-crypto/src/scrypt/fromU8a.ts`), where scrypt parameters (`N`, `p`, `r`) are extracted using hardcoded offsets without validation. This approach assumes a fixed binary layout and lacks structure or integrity checks, making it fragile and error-prone.

We currently reject any non-default values, limiting flexibility and introducing risk if malformed data is parsed. Future improvements should define a strict format and allow safe, validated parameter overrides.

### DIFF
--- a/packages/util-crypto/src/scrypt/fromU8a.ts
+++ b/packages/util-crypto/src/scrypt/fromU8a.ts
@@ -14,6 +14,11 @@ interface Result {
 }
 
 export function scryptFromU8a (data: Uint8Array): Result {
+  // Ensure the input is exactly 44 bytes: 32 for salt + 3 * 4 for N, p, r
+  if (data.length !== 32 + 12) {
+    throw new Error(`Invalid input length: expected 44 bytes, found ${data.length}`);
+  }
+
   const salt = data.subarray(0, 32);
   const N = u8aToBn(data.subarray(32 + 0, 32 + 4), BN_LE_OPTS).toNumber();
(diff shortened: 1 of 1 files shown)


packages/util-crypto/src/scrypt/fromU8a.ts
```diff
@@ -14,6 +14,15 @@ interface Result {
...
+    throw new Error('Expected input to be a Uint8Array');
+  }
+
+  // Ensure the input is exactly 44 bytes: 32 for salt + 3 * 4 for N, p, r
+  if (data.length < 32 + 12) {
+    throw new Error(`Invalid input length: expected 44 bytes, found ${data.length}`);
+  }
+
   const salt = data.subarray(0, 32);
   const N = u8aToBn(data.subarray(32 + 0, 32 + 4), BN_LE_OPTS).toNumber();
```
valentinfernandez1, line 27: Can we get this additions replaced by the actual number? there is no point on doing those operations dinamically. If we really want to make it clear why we are using those numbers we can add a comment.
```suggestion
  const N = u8aToBn(data.subarray(32, 36), BN_LE_OPTS).toNumber();
```

ap211unitech: Done

PR #2024, opened 2026-03-22
Fix(hw-ledger): reset transport on operation errors and add explicit disconnect API

F-OBrien:
## Summary
This PR updates hw-ledger transport lifecycle handling in two places:

- close/reset the active transport when a `withApp` operation fails
- add an explicit `disconnect()` method on `Ledger` and `LedgerGeneric` so consumers can clean up intentionally

## Why
While testing extension flows, I kept hitting transport/session problems after failures and across hook lifecycles:

- if an operation fails, keeping the same app/transport around can cause follow-up errors and inconsistent state
- consumers need a reliable way to release an active transport during lifecycle transitions (unmount, refresh, replacing an instance)

Related context:
- https://github.com/polkadot-js/extension/issues/1610
- https://github.com/polkadot-js/extension/pull/1611

Behavior:
- on `withApp` error: best-effort `transport.close()`, clear cached app, rethrow original error
- `disconnect()`: explicit consumer API to close an active transport without waiting for an internal failure path

## Why `disconnect()` is needed
In `useLedger`-style flows, one hook instance can unmount while its transport is still open.  
The next instance may then fail with “already open” transport errors.  
`disconnect()` gives consumers a clean, explicit teardown point to avoid that.

## Compatibility / risk
- error-path behavior is intentionally stricter: failed operations now reset transport state instead of leaving a possibly bad session alive

### DIFF
--- a/packages/hw-ledger/src/Ledger.ts
+++ b/packages/hw-ledger/src/Ledger.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { SubstrateApp } from '@zondax/ledger-substrate';
-import type { TransportDef, TransportType } from '@polkadot/hw-ledger-transports/types';
+import type { Transport, TransportDef, TransportType } from '@polkadot/hw-ledger-transports/types';
 import type { AccountOptions, LedgerAddress, LedgerSignature, LedgerVersion } from './types.js';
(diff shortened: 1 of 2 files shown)


TarikGul (approved):
👍 Looks solid! Thanks

